PR #6, opened 2020-07-08
.gitignore file with master

JoanneOFlynn2018:
.gitignore file with master.html only

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,1 @@
+master.html


cbudz:
@bergerhoffer is this sufficient or would we possibly want to add `*.html`?

bergerhoffer:
@cbudz Nope I think that master.html is fine for now. If we see the need down the line to ignore other html files, we can always update.

Thanks for submitting this @JoanneOFlynn2018!

PR #7, opened 2020-07-14
Fixing WIP notice since red doesn't render

bergerhoffer:


bergerhoffer:
@cbudz Do you mind reviewing? Just a simple change, but good practice to get into to have all changes reviewed before merging.

cbudz:
👍  change looks good to me, and I agree with getting into this practice.

PR #9, opened 2020-07-17
Add anchors to style files

cbudz:
This PR adds anchors to each file in the `style_guidelines` directory.

### DIFF
--- a/supplementary_style_guide/style_guidelines/code-commands.adoc
+++ b/supplementary_style_guide/style_guidelines/code-commands.adoc
@@ -1,3 +1,6 @@
+[discrete]
+[[code-examples]]
+
 = Code examples
 
 == Code example syntax highlighting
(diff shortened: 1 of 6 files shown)


supplementary_style_guide/style_guidelines/code-commands.adoc
```diff
@@ -1,3 +1,6 @@
+[discrete]
```
bergerhoffer, line 1: So this [discrete] tag shouldn't be here. It's actually causing build errors because this level heading isn't being shown in the TOC anymore, but the following headings are.

For now, I would remove all of the [discrete] tags that were added. We can always decide later whether the lower-level items should have them or not. But for now there are so few items that I think it's good to leave them in the TOC.

Also, all of the other headings in these files should probably get anchors as well.

bergerhoffer:
Thanks @cbudz! Merging

PR #11, opened 2020-07-21
Adding a style guideline entry

bergerhoffer:
Preview (internal): http://file.rdu.redhat.com/~ahoffer/2020/pr-11/master.html#graphical-interfaces

### DIFF
--- a/supplementary_style_guide/introduction/about-guide.adoc
+++ b/supplementary_style_guide/introduction/about-guide.adoc
@@ -5,4 +5,6 @@
 
 This guide provides style guidelines and preferred term usage for link:https://www.redhat.com/[Red Hat] product and cross-product solution documentation. It is intended to be used as a supplement to _The IBM Style Guide_, which is the primary source of guidance.
 
+// TODO: Mention that this should be referenced before the ibm style guide as it overrides some of the existing guidance
+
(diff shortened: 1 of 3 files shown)


bergerhoffer:
@cbudz I took a stab at adding an entry, if you could review and provide any feedback. This one was the UI element one, which actually had multiple parts to the recommendation, so I thought it would be best to break that out into separate guidelines.

As we discussed earlier, some of these recommendations are tied to AsciiDoc, or would be helpful to provide AsciiDoc guidance, so I've done so here. I added the caveat early on in the guide about that the examples being provided in AsciiDoc, but that the guidelines can apply to any language. We can discuss this as a team, but personally I think it's helpful, and okay to tie us to AsciiDoc (the mod docs manual does also). If we ever move away from AsciiDoc, we can always remove those examples. And anyone using the guidelines for another language can just ignore them.

As for a "template" for the entries, there's not much of one that I thought we'd need (like we have for the glossary). I figure the guidance will come in paragraph form. And then if applicable, an example AsciiDoc block. Let me know if you have any other thoughts.

bergerhoffer:
@cbudz I resolved the conflict in this PR. If you're okay with this, can you merge?

cbudz:
@bergerhoffer 👍  LGTM. Thanks for resolving that conflict. 

I agree with your assessment on implementing guidance in paragraph form and (possible) accompanying example Asciidoc block. I think providing the example is ideal, tbh.

PR #15, opened 2020-08-12
Adding guidelines for style entries

bergerhoffer:
@redhat-documentation/ccs-style-council Style entry guidance, please review! Let me know if you can think of any considerations that we need to add here.

See https://github.com/redhat-documentation/supplementary-style-guide/blob/master/supplementary_style_guide/style_guidelines/graphical-interfaces.adoc for an example implementation.

### DIFF
--- a/GUIDELINES.adoc
+++ b/GUIDELINES.adoc
@@ -23,7 +23,7 @@ Use the following template for glossary entries:
 ----
 
 [options="header"]
-|======================
+|===
 |Replace                |With                               | Notes
 |`<id>`                 |The term in lowercase.             |If the term consists of more than one word, separate them with a hyphen ("-"). Do not use spaces, capital letters, or special characters (for example, slash "/") in the ID.
 |`<term>`               |The term to be described.          |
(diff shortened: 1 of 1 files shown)


roldanbob:
Curious about the decision behind the text entry guidance. The IBM Style Guide says to use "type" to indicate that a user should provide text. I suspect that is the usage that is most common across our docs, but I haven't really investigated that hypothesis.

I was also going to say something about using `->` vs just `>` to concatenate a sequence of menu commands, but then I tested it out and saw that asciidoc converts the first one into an arrow. Cool! I never knew that.

sterobin:
@roldanbob Apparently both "enter" and "type" are permitted. See "type" under Word Usage: https://www.ibm.com/developerworks/library/styleguidelines/  I think the team decided on "enter" just because having the two options was creating inconsistency even with the same documents, and might have misled. I think the idea was to just choose one for simplicity, and "enter" was voted as most favorable by the majority. That's all, I believe.

stoobie:
> @roldanbob Apparently both "enter" and "type" are permitted, and even "issue" in the case of CLI. See "type" under Word Usage: https://www.ibm.com/developerworks/library/styleguidelines/ I think the team decided on "enter" just because having the two options was creating inconsistency even with the same documents, and might have misled. I think the idea was to just choose one for simplicity, and "enter" was voted as most favorable by the majority. That's all, I believe.
@roldanbob @sterobin **Enter** includes inputting the text and hitting the "Enter/Return" key, or whatever button might do the same thing. **Type** literally means "Hit the keys on your keyboard". It doesn't necessarily mean that you should hit the "Enter" key.

sterobin:
@stoobie Absolutely, makes sense. I think that's the simplification this change is intentionally making, i.e., not worry about "just typing" vs. "typing and submitting". Just focus on typing and submitting, which is the majority of case. I suppose we could make this distinction, that writers are welcome to use "type" when it genuinely means that thing, but again, my interpretation is to simplify for the general writer audience. Either way, I actually wasn't involved in that guidance, just interpreting, so fine by me to add the great detail there. Thanks Steve.

cbudz:
Guidelines LGTM 👍

PR #19, opened 2020-08-28
Abbreviations in titles and headings

JoanneOFlynn2018:
Hi folks,

I followed the IBM style guide as I couldn't find any other guidance in the spreadsheets or agenda. 

According to the style guide, we are telling the writer NOT to use abbreviations in headings unless it's well known or the long form is too long. In the IBM style guide, there are then 3 examples, including an example where the abbreviation is used. I included the 3 examples here but should we rank the examples, first being the most preferable?

Looking forward to your feedback!

### DIFF
--- a/supplementary_style_guide/style_guidelines/grammar.adoc
+++ b/supplementary_style_guide/style_guidelines/grammar.adoc
@@ -5,3 +5,19 @@
 
 [[abbreviations-acronyms]]
 == Abbreviations and acronyms
+
+Do not use abbreviations in headings and titles, unless an abbreviation is well known or the spelled-out form is too long. If you use an abbreviation for the first time in a heading or title, ensure that you include the spelled-out form either in the heading or title, or in the text that immediately follows the heading or title.
+
+.Examples
+
+**Creating Unified Modeling Language diagrams**
+
+You can use Unified Modeling Language (UML) diagrams to model complex systems.
+
+**Creating UML diagrams**
+
+You can use Unified Modeling Language (UML) diagrams to model complex systems.
+
+**Creating Unified Modeling Language (UML) diagrams**
+
+You can use UML diagrams to model complex systems.
--- a/supplementary_style_guide/style_guidelines/grammar.adoc
+++ b/supplementary_style_guide/style_guidelines/grammar.adoc
@@ -3,7 +3,7 @@
 = Grammar
 
 
-[[abbreviations-acronyms]]
+[id="abbreviations-acronyms"]
 == Abbreviations and acronyms
 
 Do not use abbreviations in headings and titles, unless an abbreviation is well known or the spelled-out form is too long. If you use an abbreviation for the first time in a heading or title, ensure that you include the spelled-out form either in the heading or title, or in the text that immediately follows the heading or title.


supplementary_style_guide/style_guidelines/grammar.adoc
```diff
@@ -5,3 +5,19 @@
 
 [[abbreviations-acronyms]]
```
stoobie, line 6: `[id="abbreviations-acronyms"]`

stoobie (changes requested):
@JoanneOFlynn2018 If this is just a duplicate of the topic in the IBM Style Guide, then it shouldn't be part of the supplementary style guide. Is there anything here that is unique to Red Hat?

JoanneOFlynn2018:
> @JoanneOFlynn2018 If this is just a duplicate of the topic in the IBM Style Guide, then it shouldn't be part of the supplementary style guide. Is there anything here that is unique to Red Hat?

Hi Steven, 
Thanks for your review :) It is indeed just a duplicate! However, I would rather only use the first example and omit the other 2 examples. I would also like to give an example from our own documentation. What do you think? Worth changing and keeping, or delete this section outright?

stoobie:
> > @JoanneOFlynn2018 If this is just a duplicate of the topic in the IBM Style Guide, then it shouldn't be part of the supplementary style guide. Is there anything here that is unique to Red Hat?
> 
> Hi Steven,
> Thanks for your review :) It is indeed just a duplicate! However, I would rather only use the first example and omit the other 2 examples. I would also like to give an example from our own documentation. What do you think? Worth changing and keeping, or delete this section outright?

Hi @JoanneOFlynn2018, so are you saying that the Red Hat style guide,doesn't allow using acronyms/abbreviations in headings at all?

supplementary_style_guide/style_guidelines/grammar.adoc
```diff
@@ -3,5 +3,21 @@
 = Grammar
 
 
-[[abbreviations-acronyms]]
+[id="abbreviations-acronyms"]
 == Abbreviations and acronyms
+
+Do not use abbreviations in headings and titles, unless an abbreviation is well known or the spelled-out form is too long. If you use an abbreviation for the first time in a heading or title, ensure that you include the spelled-out form either in the heading or title, or in the text that immediately follows the heading or title.
```
bergerhoffer, line 9: Yeah, as @stoobie pointed it, it does seem like this guidance is an exact copy of IBM's, which we don't want to duplicate here.

From the spreadsheet, the recommended guidance is in this cell here: https://docs.google.com/spreadsheets/d/1BBgV0FV72LIIh0e7Bw9yPKN4I_ikYtTWREb7thT2cyA/edit#gid=2022541005&range=I10

In looking at the guidance there again, it really is pretty much just saying to follow IBM's guidance. The only thing that I can think of, is do we want to make any sort of recommendation on headings in modules vs assemblies. For example, make sure that the assembly heading always spells it out, but modules are okay to use abbreviations in their heading (while still spelling it out in the text following the heading). If we don't want to make any sort of mention or exceptions like that, then I'd say that this shouldn't be an item in our guide after all.

Thoughts @redhat-documentation/ccs-style-council?

JoanneOFlynn2018: I think some guidance on use of abbreviations in module headings vs assembly headings is a good idea! 
What does everyone else think? 
@redhat-documentation/ccs-style-council

leswilliams44: I like the idea of spelling out an acronym in an assembly and using an acronym in modules. I think we should stay away from using abbreviations.

IngridT1: Would the guidance be any different than the IBM guidance? Or would we want to add additional nuance of some kind? @JoanneOFlynn2018 

I agree with @bergerhoffer if it's just an exact repeat, we should leave it out.

bergerhoffer: @redhat-documentation/ccs-style-council Should we discuss this at next week's meeting?

I see two options on how to move forward with this:

- Defer completely to IBM's guidance and don't include an entry in our guide for acronyms.
- Add an entry to our guide that says to follow IBM's guidance on acronyms, but specifically make a recommendation to spell out acronyms in assembly headings. Caveat: not ALL acronyms; we don't want to spell out well-known acronyms (like IBM gives an example of HTML for this).

JoanneOFlynn2018: @bergerhoffer I would like to discuss this next week Andrea, thanks for the suggestion!

bergerhoffer: Thanks @JoanneOFlynn2018! I've added this to the agenda.

cbudz: @bergerhoffer I'd like to see what the possible implications are for assemblies/modules moving to Pantheon v2 since that will support module-level publishing. Would like to hear Emily or Stetson weigh in from an FCC perspective.

bergerhoffer: Good question

cbudz:
Per the 929 meeting of the Assembly Templates Committee, the Pantheon v2 will research the possibility for using variables in titles at the assembly and module level, post milestone 3. At the moment variables in assembly/module titles are not supported. 

I suggest we leave this recommendation `on hold` for the near future depending on what that team comes back with.

supplementary_style_guide/style_guidelines/grammar.adoc
sterobin: @cbudz @bergerhoffer, just chiming in in response to a ping from @gregraka to me and @emmurphy1  about this convo. IMO, I think the current guidance here described, and in the IBM SG, still should be followed even in a Pv2 standalone-module scenario, and thus can probably just be dropped from our own SG if that's the group feeling. Reason is that it's perfectly reasonable to abbreviate obvious things and extremely lengthy things, typically our own product names, to make the doc structure more penetrable. I think the key is to keep in mind the target audience when deciding what's "obvious" and "lengthy" vs. what's truly needed. The argument about standalone modules is accounted for by the fact that IBM and we still prescribe that if you do abbreviate in the title, ensure that upon first mention in the following paragraph that it's expanded in full form. That takes care of it IMO. In some cases that too might not even be necessary depending, but for the prescription, a good side to err on. 

So I think it's a win-win-win to go the current route of abbreviations in titles as in IBM, and can likely be dropped from our SG: win for simplicity, for IBM adherence, and for still accommodating standalone modules. Just my thoughts.

@emmurphy1 Any other thoughts or objections?

emmurphy1: I agree with @sterobin and this is the guidance I have been giving when it comes  up in the context of modularization.

gregraka:
Hi,

Based on Stetson and Emily's latest comments, "...to go the current route of abbreviations in titles as in IBM, and can likely be dropped from our SG."

Joanne (@JoanneOFlynn2018), if you disagree, please re-open this issue.

Thanks,
--Greg

PR #27, opened 2020-09-29
Add guidelines about using attributes for product names and versions

inoxx03:
@bergerhoffer @cbudz @redhat-documentation/ccs-style-council 
I have added initial content for the guidelines on using product name and version attributes.
The additions that I made so far are very general and should be applicable to all project.
However I do feel that more details are necessary and I would like to know you input on the current content before I proceed with adding more of it.

Note, that this is still a work in progress and you can find some of my proposed content in the comments. this content is not complete and will have to be further developed.

Please let me know your thoughts and comments. Thank you.

### DIFF
--- a/supplementary_style_guide/style_guidelines/general-formatting.adoc
+++ b/supplementary_style_guide/style_guidelines/general-formatting.adoc
@@ -109,13 +109,13 @@ Only use hard-coded version references if the version that you are referring to 
 === Attribute file
 
 Define attributes for product name and product version and store them in a dedicated attributes file for each set of product documentation.
-For examples of where you can store the shared attributes file inside your documentation repository, see the link:https://github.com/redhat-documentation/modular-docs/tree/mod-doc-repo-example[Example modular documentation repository].
+For examples of where you can store the shared attributes file inside your documentation repository, see the link:https://github.com/redhat-documentation/modular-docs/blob/mod-doc-repo-example/_artifacts/document-attributes.adoc[Example modular documentation repository].
 Include the attributes file at the beginning of the `master.adoc` files of all titles in your documentation set:
 
-.master.adoc
-[source,asciidoc,options="nowrap"]
+.Example Asciidoc: Attribute file included in a master.adoc file
+[subs="+quotes"]
 ----
-\include::<path-to-directory-with-attributes-file>/attributes.adoc[]
+\include::__<path_to_directory_with_attributes_file>__/attributes.adoc[]
 ----
 
 === Minimum required attributes
@@ -128,25 +128,25 @@ The name of the product:: Use the product name attribute for all instances of th
 Avoid using hard-coded product names.
 For example:
 +
-.Example product name attribute
+.Example Asciidoc: Product name attribute
 ----
 :name-product: Red Hat build of Quarkus
 ----
 
 The abbreviated form of the product name:: If it is necessary for your product, you can use an attribute to store a shortened version of the name of your product, for example:
 +
-.Example abbreviated product name attribute
+.Example Asciidoc: Abbreviated product name attribute
 ----
-:name-product-abbreviated: RHBQ
+:name-product-abbreviated: JBoss EAP for OpenShift
 ----
 
 The major and minor version of the product:: Use an attribute for the product version in cases where the product version can change with each release and the content is still correct.
 For example:
 +
-.Example long and short product version attributes
+.Example Asciidoc: Long and short product version attributes
 ----
-:version-product-short: 1.11.6
-:version-product-long: 1.11
+:version-product-minor: 1.11
+:version-product-patch: 1.11.6
 ----
 +
 [NOTE]
--- a/supplementary_style_guide/style_guidelines/general-formatting.adoc
+++ b/supplementary_style_guide/style_guidelines/general-formatting.adoc
@@ -103,71 +103,64 @@ Text for note.
 [product-names-versions-ref]]
 == Product names and version references
 
-////
-PROPOSED ADDITIONS?
-master-docinfo.xml specific and portal-specific version variables
-URL fragments and composed links?
-correct sequence when combining attributes
-ensure proper sequence for compound attributes
-syntax recommendations: what separators to use? dashes/hyphens, underscores?
-////
-
 Whenever you refer to the name of your product in full, or in its abbreviated form, or when you refer to the major and minor version of your product, avoid using hard-coded references and use attributes instead.
-
 Only use hard-coded version references if the version that you are referring to in a particular case never changes.
 
 === Attribute file
 
-Define attributes for product name and product version in a single common shared attributes file for each documentation set or product.
-
-Name the file `attributes.adoc`.
-You may use a different name if necessary, but if you do, ensure that the name of the file indicates to your collaborators that it is an attribute file.
-
-//For a single title, store the file in the same directory as the `master.adoc` file.
-For sharing between multiple titles, store the file in a common directory above the directories that contain your titles, such as `docs/common/` or `docs/topics/templates`.
-
+Define attributes for product name and product version and store them in a dedicated attributes file for each set of product documentation.
+For examples of where you can store the shared attributes file inside your documentation repository, see the link:https://github.com/redhat-documentation/modular-docs/tree/mod-doc-repo-example[Example modular documentation repository].
 Include the attributes file at the beginning of the `master.adoc` files of all titles in your documentation set:
 
 .master.adoc
 [source,asciidoc,options="nowrap"]
 ----
-\include::common/attributes.adoc[]
+\include::<path-to-directory-with-attributes-file>/attributes.adoc[]
 ----
-
-TIP: Create a symbolic link pointing to the `common` directory inside the directory containing the `master.adoc` file of your title. It makes including the attribute file in your title easier. +
-Do not copy the directory or attribute file directly, and do not use hard links!
 
 === Minimum required attributes
 
-For each title ensure that you define attributes for the following values:
+Define attributes for the following values in each documentation set.
+Note that the attribute names used in this section are only meant as examples.
+You can use different attribute names:
 
-* product name
-* abbreviated product name (if applicable to your product)
-* the major and minor version of the product (for example, 5 and 5.4)
-
-You may create additional attributes according to what your documentation requires.
-For example, some projects with complex names may define multiple abbreviated and non-abbreviated forms for their product name. You may also combine existing product name attributes to create compound product names.
-
-=== Attribute usage
-
-Product names and abbreviations::
-Use the product name attribute for all instances of the product name where possible.
+The name of the product:: Use the product name attribute for all instances of the product name where possible.
 Avoid using hard-coded product names.
+For example:
++
+.Example product name attribute
+----
+:name-product: Red Hat build of Quarkus
+----
 
-Product versions::
-Use the product version attribute for instances where the product version should change with the next version and the content would still be correct.
+The abbreviated form of the product name:: If it is necessary for your product, you can use an attribute to store a shortened version of the name of your product, for example:
++
+.Example abbreviated product name attribute
+----
+:name-product-abbreviated: RHBQ
+----
+
+The major and minor version of the product:: Use an attribute for the product version in cases where the product version can change with each release and the content is still correct.
+For example:
++
+.Example long and short product version attributes
+----
+:version-product-short: 1.11.6
+:version-product-long: 1.11
+----
++
+[NOTE]
+====
 Do not use the product version attribute if the version should not change.
 For example, if a feature was introduced in a certain version, the version should be hard-coded.
+====
 
-////
-=== Recommendations on attribute name syntax
+You might create additional attributes according to what your documentation requires.
+For example, you might combine existing product name attributes to create compound names of products or components:
 
-The particular format of attribute names is largely determined by the internal naming conventions of each particular project.
-
-Some recommendation can be made to the general syntax of attribute names to ensure that the names clearly indicate the purpose of each attribute.
-
-Attribute names should be as explicit as possible, especially in sets that have many attributes.
-Using prefixes helps achieve this without making the attribute name too long.
-For example `name-product-short` clearly indicates that the attribute resolves.
-It is also not product specific, which might be usitable for
-////
+.Example attributes for compound names of product components
+----
+:name-runtime-spring-boot: Spring Boot
+:name-runtime-vertx: Eclipse Vert.x
+:name-spring-reactive: {name-runtime-spring-boot} with {name-runtime-vertx} reactive components
+----
--- a/supplementary_style_guide/style_guidelines/general-formatting.adoc
+++ b/supplementary_style_guide/style_guidelines/general-formatting.adoc
@@ -112,7 +112,7 @@ Define attributes for product name and product version and store them in a dedic
 For examples of where you can store the shared attributes file inside your documentation repository, see the link:https://github.com/redhat-documentation/modular-docs/blob/mod-doc-repo-example/_artifacts/document-attributes.adoc[Example modular documentation repository].
 Include the attributes file at the beginning of the `master.adoc` files of all titles in your documentation set:
 
-.Example Asciidoc: Attribute file included in a master.adoc file
+.Example AsciiDoc: Attribute file included in a master.adoc file
 [subs="+quotes"]
 ----
 \include::__<path_to_directory_with_attributes_file>__/attributes.adoc[]
@@ -128,14 +128,14 @@ The name of the product:: Use the product name attribute for all instances of th
 Avoid using hard-coded product names.
 For example:
 +
-.Example Asciidoc: Product name attribute
+.Example AsciiDoc: Product name attribute
 ----
 :name-product: Red Hat build of Quarkus
 ----
 
 The abbreviated form of the product name:: If it is necessary for your product, you can use an attribute to store a shortened version of the name of your product, for example:
 +
-.Example Asciidoc: Abbreviated product name attribute
+.Example AsciiDoc: Abbreviated product name attribute
 ----
 :name-product-abbreviated: JBoss EAP for OpenShift
 ----
@@ -143,7 +143,7 @@ The abbreviated form of the product name:: If it is necessary for your product, 
 The major and minor version of the product:: Use an attribute for the product version in cases where the product version can change with each release and the content is still correct.
 For example:
 +
-.Example Asciidoc: Long and short product version attributes
+.Example AsciiDoc: Product version attributes
 ----
 :version-product-minor: 1.11
 :version-product-patch: 1.11.6
--- a/supplementary_style_guide/style_guidelines/general-formatting.adoc
+++ b/supplementary_style_guide/style_guidelines/general-formatting.adoc
@@ -100,5 +100,74 @@ Text for note.
 ====
 ----
 
-// [[product-names-versions-ref]]
-// == Product names and version references
+[product-names-versions-ref]]
+== Product names and version references
+
+////
+PROPOSED ADDITIONS?
+master-docinfo.xml specific and portal-specific version variables
+URL fragments and composed links?
+correct sequence when combining attributes
+ensure proper sequence for compound attributes
+syntax recommendations: what separators to use? dashes/hyphens, underscores?
+////
+
+Whenever you refer to the name of your product in full, or in its abbreviated form, or when you refer to the major and minor version of your product, avoid using hard-coded references and use attributes instead.
+
+Only use hard-coded version references if the version that you are referring to in a particular case never changes.
+
+=== Attribute file
+
+Define attributes for product name and product version in a single common shared attributes file for each documentation set or product.
+
+Name the file `attributes.adoc`.
+You may use a different name if necessary, but if you do, ensure that the name of the file indicates to your collaborators that it is an attribute file.
+
+//For a single title, store the file in the same directory as the `master.adoc` file.
+For sharing between multiple titles, store the file in a common directory above the directories that contain your titles, such as `docs/common/` or `docs/topics/templates`.
+
+Include the attributes file at the beginning of the `master.adoc` files of all titles in your documentation set:
+
+.master.adoc
+[source,asciidoc,options="nowrap"]
+----
+\include::common/attributes.adoc[]
+----
+
+TIP: Create a symbolic link pointing to the `common` directory inside the directory containing the `master.adoc` file of your title. It makes including the attribute file in your title easier. +
+Do not copy the directory or attribute file directly, and do not use hard links!
+
+=== Minimum required attributes
+
+For each title ensure that you define attributes for the following values:
+
+* product name
+* abbreviated product name (if applicable to your product)
+* the major and minor version of the product (for example, 5 and 5.4)
+
+You may create additional attributes according to what your documentation requires.
+For example, some projects with complex names may define multiple abbreviated and non-abbreviated forms for their product name. You may also combine existing product name attributes to create compound product names.
+
+=== Attribute usage
+
+Product names and abbreviations::
+Use the product name attribute for all instances of the product name where possible.
+Avoid using hard-coded product names.
+
+Product versions::
+Use the product version attribute for instances where the product version should change with the next version and the content would still be correct.
+Do not use the product version attribute if the version should not change.
+For example, if a feature was introduced in a certain version, the version should be hard-coded.
+
+////
+=== Recommendations on attribute name syntax
+
+The particular format of attribute names is largely determined by the internal naming conventions of each particular project.
+
+Some recommendation can be made to the general syntax of attribute names to ensure that the names clearly indicate the purpose of each attribute.
+
+Attribute names should be as explicit as possible, especially in sets that have many attributes.
+Using prefixes helps achieve this without making the attribute name too long.
+For example `name-product-short` clearly indicates that the attribute resolves.
+It is also not product specific, which might be usitable for
+////
--- a/supplementary_style_guide/style_guidelines/general-formatting.adoc
+++ b/supplementary_style_guide/style_guidelines/general-formatting.adoc
@@ -130,14 +130,14 @@ For example:
 +
 .Example AsciiDoc: Product name attribute
 ----
-:name-product: Red Hat build of Quarkus
+:name-product: Red Hat JBoss Enterprise Application Platform
 ----
 
 The abbreviated form of the product name:: If it is necessary for your product, you can use an attribute to store a shortened version of the name of your product, for example:
 +
 .Example AsciiDoc: Abbreviated product name attribute
 ----
-:name-product-abbreviated: JBoss EAP for OpenShift
+:name-product-abbreviated: JBoss EAP
 ----
 
 The major and minor version of the product:: Use an attribute for the product version in cases where the product version can change with each release and the content is still correct.


supplementary_style_guide/style_guidelines/general-formatting.adoc
```diff
@@ -21,3 +21,72 @@ When the format _day Month year_ causes a presentation or clarity issue, use ins
...
+
+Define attributes for product name and product version in a single common shared attributes file for each documentation set or product.
+
+Name the file `attributes.adoc`.
+You may use a different name if necessary, but if you do, ensure that the name of the file indicates to your collaborators that it is an attribute file.
+
+//For a single title, store the file in the same directory as the `master.adoc` file.
+For sharing between multiple titles, store the file in a common directory above the directories that contain your titles, such as `docs/common/` or `docs/topics/templates`.
+
+Include this file at the beginning of the `master.adoc` files of all titles in your documentation set:
```
ncbaratta, line 48: 'this' was making me look back to find out what it was referring to:

```suggestion
Include the attributes file at the beginning of the `master.adoc` files of all titles in your documentation set:
```

supplementary_style_guide/style_guidelines/general-formatting.adoc
```diff
@@ -21,3 +21,72 @@ When the format _day Month year_ causes a presentation or clarity issue, use ins
...
+
+.master.adoc
+[source,asciidoc,options="nowrap"]
+----
+\include::common/attributes.adoc[]
+----
+
+TIP: Create a symbolic link pointing to the common directory inside the directory containing the `master.adoc` file of your title. It makes including the attribute file in your title easier. +
```
ncbaratta, line 56:
```suggestion
TIP: Create a symbolic link pointing to the `common` directory inside the directory containing the `master.adoc` file of your title. It makes including the attribute file in your title easier. +
```

ncbaratta (dismissed):
LGTM

supplementary_style_guide/style_guidelines/general-formatting.adoc
```diff
@@ -21,3 +21,72 @@ When the format _day Month year_ causes a presentation or clarity issue, use ins
...
+correct sequence when combining attributes
+ensure proper sequence for compound attributes
+syntax recommendations: what separators to use? dashes/hyphens, underscores?
+////
+
+Whenever you refer to the name of your product in full, or in its abbreviated form, or when you refer to the major and minor version of your product, avoid using hard-coded references and use attributes instead.
+
+Only use hard-coded version references if the version that you are referring to in that particular case never changes.  
+
+=== Attribute file
```
stoobie, line 38: @inoxx03 I thought that in mod docs there's a guideline about starting every module with a heading 1 and using offsets in the assembly to define the nesting.

supplementary_style_guide/style_guidelines/general-formatting.adoc
```diff
@@ -21,3 +21,72 @@ When the format _day Month year_ causes a presentation or clarity issue, use ins
...
+master-docinfo.xml specific and portal-specific version variables
+URL fragments and composed links?
+correct sequence when combining attributes
+ensure proper sequence for compound attributes
+syntax recommendations: what separators to use? dashes/hyphens, underscores?
+////
+
+Whenever you refer to the name of your product in full, or in its abbreviated form, or when you refer to the major and minor version of your product, avoid using hard-coded references and use attributes instead.
+
+Only use hard-coded version references if the version that you are referring to in that particular case never changes.
```
stoobie, line 36: Suggestion: `in that particular case` -> `in a particular case`

supplementary_style_guide/style_guidelines/general-formatting.adoc
```diff
@@ -21,3 +21,72 @@ When the format _day Month year_ causes a presentation or clarity issue, use ins
...
+
+Include the attributes file at the beginning of the `master.adoc` files of all titles in your documentation set:
+
+.master.adoc
+[source,asciidoc,options="nowrap"]
+----
+\include::common/attributes.adoc[]
+----
+
+TIP: Create a symbolic link pointing to the `common` directory inside the directory containing the `master.adoc` file of your title. It makes including the attribute file in your title easier. +
```
stoobie, line 56: This assumes a certain directory structure. I don't think that this is the place to make that assumption. I recall that @sterobin leads a class in setting up directory structure in which this tip would not make sense. Also, I believe that using symlinks to the /common directory is only necessary when a certain Asciidoctor flag is set, which has been the case in the Pantheon toolchain. Not sure that it will continue to be so in Pantheon 2, and I'm also not sure that upstream projects need to worry about this. I know that with oVirt docs, we need to change references to the symlinked /common directory before publishing to upstream.

supplementary_style_guide/style_guidelines/general-formatting.adoc
```diff
@@ -21,3 +21,72 @@ When the format _day Month year_ causes a presentation or clarity issue, use ins
...
 [[product-names-versions-ref]]
 == Product names and version references
+
+////
+PROPOSED ADDITIONS?
+master-docinfo.xml specific and portal-specific version variables
+URL fragments and composed links?
+correct sequence when combining attributes
+ensure proper sequence for compound attributes
+syntax recommendations: what separators to use? dashes/hyphens, underscores?
```
stoobie, line 31: Do mod docs guidelines provide guidance on this? @sterobin ?

supplementary_style_guide/style_guidelines/general-formatting.adoc
```diff
@@ -21,3 +21,72 @@ When the format _day Month year_ causes a presentation or clarity issue, use ins
 
 [[product-names-versions-ref]]
 == Product names and version references
```
stoobie, line 23: @bergerhoffer @cbudz @inoxx03 :
This entire topic is about Asciidoc implementation guidelines, not about style guidelines. I thought that we agreed as a group that the style guide would provide Asciidoc examples, but I didn't think that entire Asciidoc implementation modules are within the scope of this guide.

Was there a decision made that I'm not aware of? Seems like this should be part of the Asciidoc Quick Reference that we talked about last week.

supplementary_style_guide/style_guidelines/general-formatting.adoc
```diff
@@ -21,3 +21,72 @@ When the format _day Month year_ causes a presentation or clarity issue, use ins
...
+=== Minimum required attributes
+
+For each title ensure that you define attributes for the following values:
+
+* product name
+* abbreviated product name (if applicable to your product)
+* the major and minor version of the product (for example, 5 and 5.4)
+
+You may create additional attributes according to what your documentation requires.
+For example, some projects with complex names may define multiple abbreviated and non-abbreviated forms for their product name. You may also combine existing product name attributes to create compound product names.
```
stoobie, line 68: Can you give an example?

supplementary_style_guide/style_guidelines/general-formatting.adoc
```diff
@@ -21,3 +21,72 @@ When the format _day Month year_ causes a presentation or clarity issue, use ins
...
+----
+
+TIP: Create a symbolic link pointing to the `common` directory inside the directory containing the `master.adoc` file of your title. It makes including the attribute file in your title easier. +
+Do not copy the directory or attribute file directly, and do not use hard links!
+
+=== Minimum required attributes
+
+For each title ensure that you define attributes for the following values:
+
+* product name
```
stoobie, line 63: I think it would be clearer to make this a description list and use the actual syntax. If you're worried about teams using differing syntax, you can add a note to that effect before or after the list. You could also make the topic more concise that way. In fact, I think you could combine the _Minimum required attributes_ and _Attribute usage_ in one section.

stoobie (changes requested):

supplementary_style_guide/style_guidelines/general-formatting.adoc
```diff
@@ -21,3 +21,72 @@ When the format _day Month year_ causes a presentation or clarity issue, use ins
...
+
+=== Attribute file
+
+Define attributes for product name and product version in a single common shared attributes file for each documentation set or product.
+
+Name the file `attributes.adoc`.
+You may use a different name if necessary, but if you do, ensure that the name of the file indicates to your collaborators that it is an attribute file.
+
+//For a single title, store the file in the same directory as the `master.adoc` file.
+For sharing between multiple titles, store the file in a common directory above the directories that contain your titles, such as `docs/common/` or `docs/topics/templates`.
```
pwright, line 46: `docs/artifacts` might be a better example

supplementary_style_guide/style_guidelines/general-formatting.adoc
```diff
@@ -21,3 +21,72 @@ When the format _day Month year_ causes a presentation or clarity issue, use ins
...
+Include the attributes file at the beginning of the `master.adoc` files of all titles in your documentation set:
+
+.master.adoc
+[source,asciidoc,options="nowrap"]
+----
+\include::common/attributes.adoc[]
+----
+
+TIP: Create a symbolic link pointing to the `common` directory inside the directory containing the `master.adoc` file of your title. It makes including the attribute file in your title easier. +
+Do not copy the directory or attribute file directly, and do not use hard links!
```
pwright, line 57: !

supplementary_style_guide/style_guidelines/general-formatting.adoc
inoxx03: @inoxx03 You are absolutely correct! But ironically, the Supplementary Style Guide document is not (yet) modular.  It's still a monolith (I was surprised to find out about that after having cloned the repo). That being said, I'd be happy to lead by example and change my chapter into a properly chunked and formatted module. However we should all adopt that approach in our individual contributions and, going forward, also make it an action item to retroactively modularize the existing monolithic sections.

oraNod (commented):
@inoxx03 I'm not sure if this is overkill or not. Not to rain on anyone's parade but I don't get how attributes files fall under modular documentation. And I don't agree that we should be mandating any rules about defining attributes.

supplementary_style_guide/style_guidelines/general-formatting.adoc
bergerhoffer: @stoobie yeah for now, we didn't follow the mod docs layout because we ported what we had from the glossary of terms and conventions in here. I'm definitely open to tackling it at some point, but I didn't think it was worthwhile to worry about transforming all of that right now.

ncbaratta:
> @inoxx03 I'm not sure if this is overkill or not. Not to rain on anyone's parade but I don't get how attributes files fall under modular documentation. And I don't agree that we should be mandating any rules about defining attributes.

We have to start mandating some rules around attributes because one part of modular docs is sharing content and that means we'll be affected by other products' attributes.

supplementary_style_guide/style_guidelines/general-formatting.adoc
bergerhoffer: @stoobie So this has been an item on the list of recommendations since day 1. I agree that it crosses a bit into implementation, as did the technology preview item. We did discuss the TP one at one of our meetings a month or two back. (it was in this list of items that weren't okayed by teams when we sent them out: https://docs.google.com/document/d/1rKTZXqjS8QJZ-gPpXp-UOFLYaX8IJKuQl2Pzxr-rQwQ/edit). The consensus there was to potentially move it to more of an implementation-specific section. The same could potentially be considered for this item.

However, @inoxx03, the guidance you're adding here goes a lot further than the initial recommendation that we sent out. This is what was in the spreadsheet for this item:

```
Define attributes for product name and product version in a common attributes file, to be included in all books/master.adoc files.

Use the product name attribute for all instances of the product name where possible. Some products define more than one attribute, for example, if there is a full name with an abbreviation, and a short name of just the abbreviation.

Use the product version attribute for instances where the product version should change with the next version and the content would still be correct. Do not use the product version attribute if the version should not change. For example, if a feature was introduced in a certain version, the version should be hard coded.
```

It's moreso general tips/suggestions for the best practice of using attributes for these items. And less about the actual implementation of what to name the file, where to put it, etc.

I'm torn on what to do about this one. I think that it's a good idea to recommend using attributes for things like product name and versions, but I'd be hesitant to include specifics without coordinating with the FCC group to make sure that what we recommend is going to work when sharing content across products.

@redhat-documentation/ccs-style-council Thoughts?

oraNod:
> > @inoxx03 I'm not sure if this is overkill or not. Not to rain on anyone's parade but I don't get how attributes files fall under modular documentation. And I don't agree that we should be mandating any rules about defining attributes.
> 
> We have to start mandating some rules around attributes because one part of modular docs is sharing content and that means we'll be affected by other products' attributes.

Yeah I get the sharing aspect but all these wordy rules... Why don't we just provide some example attributes.adoc files with "recommended" variables such as :ProductName: or whatever? No offence to anyone working on it but these guidelines read like we're stating the obvious. I'd imagine most doc sets already use attributes.

IMO it's worthwhile to consider a strategy for dealing with attributes in various products. With DG, for example, we inherited :brandname: from the community docs as we're sourced upstream. So we use that instead of :ProductName:

I considered a simple substitution during the downstream sync job `s/brandname/ProductName` but figured the path of least resistance would be to just use the community attribute.

We're going to have to deal with conflicting and inconsistent attributes across doc sets. Attributes are difficult to change once they are defined and part of the doc set. Going back to the DG :brandname: attribute, I didn't ask the team to change that to ProductName because they're familiar with it and used to it when writing docs. 

Granted it is better to have standards and common attributes to avoid having to do substitutions and stuff like that. It's kind of like all the different phone numbers in Germany. Some parts of the country have 5 digits, other parts 7 digits, and it varies pretty wildly across regions. There are all these esoteric lookup rules because formats are so different. It's a mess.

So, yeah, we want to avoid a similar mess but I feel like we shouldn't be trying to make things too homogeneous at the same time. Hopefully this makes some sense.

pwright:
@inoxx03  I agree with @oraNod  that there may be scenarios where these guidelines are not adhered to, but also see the value for new writers/projects. Maybe some caveat in the introduction could resolve this, something like 'Typically, RH sponsored projects use these guidelines, but some communities might have their own conventions.'

oraNod:
> @inoxx03 I agree with @oraNod that there may be scenarios where these guidelines are not adhered to, but also see the value for new writers/projects. Maybe some caveat in the introduction could resolve this, something like 'Typically, RH sponsored projects use these guidelines, but some communities might have their own conventions.'

Again, that seems like just stating the obvious. If this is aimed for new projects, why don't we just provide an attributes file with some recommended attributes to get going? 

Thinking back to Nicole's point about sharing attributes for cross-product docs, have we considered how to resolve conflicting attributes? For example, RHDG doc set uses ProductName and RHSSO doc set uses ProductName. When we start combining modules from both doc sets, then it will make sense for ProductName to resolve to RHDG in some instances and RHSSO in other instances. 

We're going to run into scenarios like that no matter what guidelines or rules are in place for defining attributes.

pwright:
@oraNod re. assembly with modules that define ProductName differently.
 isn't that where 'glue text' helps?

```
:ProductName: my product
include::my-module[]
:ProductName: your product
include::your-module[]
```

oraNod:
@pwright yeah that could help with modules that define ProductName differently. although I could see that being problematic as it moves attribute definitions out of the attributes file and into the assembly, which could lead to "spaghetti code" that is a nightmare to maintain.

supplementary_style_guide/style_guidelines/general-formatting.adoc
inoxx03:
```suggestion
Only use hard-coded version references if the version that you are referring to in a particular case never changes.  
```

supplementary_style_guide/style_guidelines/general-formatting.adoc
inoxx03: Thanks for pointing that out. Yes, I'll add an example of what I'm describing (unless we remove the section of text in question).

supplementary_style_guide/style_guidelines/general-formatting.adoc
inoxx03: @pwright Thank you. One question: Better than which one of the two that I put there? :)
While both directory path examples I included here are based on examples taken from EAP, Quarkus, CNR and JWS, I am unaware of any project at red hat that stores attribute files in `docs/artifacts`. (You can see that I fell victim to my own bias based on experience.) 
That being said I believe that `docs/common` is as valid an example as any here (Especially considering that `attributes.adoc` is in its nature and intended purpose neither an artifact nor a template).
I will add `docs/artifacts` as another example of a location where `attributes.adoc` can be stored.

supplementary_style_guide/style_guidelines/general-formatting.adoc
inoxx03: I take it that exclamation marks are a stylistic no-no. Understandable. I'll get rid of it.

supplementary_style_guide/style_guidelines/general-formatting.adoc
inoxx03: Another alternative that comes to mind is to avoid giving an example altogether. Writers can be reasonably expected to comprehend the guideline based on what is written down, even without an example.That way we can avoid sounding too prescriptive if any of you are concerned about that. Thoughts?

pwright: @inoxx03  the reason I suggested this is because that's the standard ;) https://github.com/redhat-documentation/modular-docs/tree/mod-doc-repo-example
But, of course, because I'm not using that standard, I got the example wrong... should have been `docs/_artifacts` 

I think leaving `docs/common` there makes sense, cos it reflects common usage.

supplementary_style_guide/style_guidelines/general-formatting.adoc
inoxx03:
```suggestion
Do not copy the directory or attribute file directly, and do not use hard links.
```

supplementary_style_guide/style_guidelines/general-formatting.adoc
stoobie: @inoxx03 I agree with @pwright about leaving it as-is, but just make it clear that this is an example.

supplementary_style_guide/style_guidelines/general-formatting.adoc
inoxx03: @bergerhoffer My initial intention was to provide some context and additional information around the basic recommendations that have been noted down in the spreadsheet.
I see that that might not be the best solution.
Please take note of me repeatedly stating in the text that the writer may use a different attribute naming scheme, file location etc, in their particular case if need be. I surmised that it would be enough to hedge against the possibility of sounding too prescriptive.  However, based on the comments that you and other reviewers have kindly provided, I see that I may have been wrong. Would you prefer just having the basic recommendation from the spreadsheet in the chapter for the time being? We can add other recommendations later, once we have reached a consensus of what said recommendations should be. Thoughs? Comments? @redhat-documentation/ccs-style-council

supplementary_style_guide/style_guidelines/general-formatting.adoc
inoxx03: @pwright Ah, thank you for the clarification:) I did not check the directory recommendation in the mod-docs guide, which is my mistake. I should have done so. Indeed, `docs/_artifacts` seems to me like a directory compatible with Pantheon MS3. Given that we want to recommend a directory structure compatible with the next generation of our publishing tools, using your suggested example seems highly appropriate. I'll add that information.

@stoobie Thank you, I will make it clear in the section that what we list as directory names is to be seen as a recommendation rather than a hard-and-fast rule.

supplementary_style_guide/style_guidelines/general-formatting.adoc
inoxx03: @stoobie You make a good point. I see at least 2 possible solutions here: 
1. Make it clear that what whatever information we provide around directory structure is only an example and primarily relevant to Red Hat projects.
2. Remove all content that deals with the details of directory structure form this chapter, leaving it entirely up to each project's internal rules what directory structure to use.
Any of these seem favorable to you? Would you have another alternative that I have not listed above?

supplementary_style_guide/style_guidelines/general-formatting.adoc
inoxx03: Thanks, @stoobie that could work.
Originally, I stayed away from specific examples precisely because I wanted to avoid inflaming a dispute about whether I am recommending a specific syntax for these attributes or not. 
In light of the comments that I have seen from other reviewers, it might be best to not recommend a specific syntax for attributes in this guide at all.
Would it make sense to have the recommended usage of attributes (incl. file name, location, etc, ...) covered in another guide (like the mod docs guidelines?) That way we can cover some of the interesting cases related to reuse such as `ProductName` resolving to different product names while used in the same title.

bergerhoffer:
I'm going to mark this one as on hold for now. We need to look into working with FCC/mod docs to compile guidance on what will work for these attributes when sharing content between products. We can revisit after the first version of the style guide goes out.

supplementary_style_guide/style_guidelines/general-formatting.adoc
stoobie: @inoxx03 Those are both good ideas. I think it would be helpful to provide a link to the mod docs guide, if that is permissible. I think that the mod docs guide might talk about dir structure.

supplementary_style_guide/style_guidelines/general-formatting.adoc
stoobie: @inoxx03 +1

supplementary_style_guide/style_guidelines/general-formatting.adoc
```diff
@@ -21,3 +21,72 @@ When the format _day Month year_ causes a presentation or clarity issue, use ins
...
+Whenever you refer to the name of your product in full, or in its abbreviated form, or when you refer to the major and minor version of your product, avoid using hard-coded references and use attributes instead.
+
+Only use hard-coded version references if the version that you are referring to in a particular case never changes.  
+
+=== Attribute file
+
+Define attributes for product name and product version in a single common shared attributes file for each documentation set or product.
+
+Name the file `attributes.adoc`.
+You may use a different name if necessary, but if you do, ensure that the name of the file indicates to your collaborators that it is an attribute file.
```
stoobie, line 43: @inoxx03 Don't use "may" in technical docs. It's ambiguous. Use "might" or "can", or any other word that is less ambiguous. In this case: `You can use a different name if necessary...`

supplementary_style_guide/style_guidelines/general-formatting.adoc
gregraka: Hi,
I can see @stoobie 's point-of-view about this content being more about implementation. However, I also get @inoxx03 's comment that product names and version attributes is a gray area (between style and implementation-FCC-asciidoc). When discussing product names and versions, I find it difficult to avoid discussing attributes. Let's include attributes and state the broadly accepted principles using them is a best practice that we encourage and support. The implementation details will work themselves out.

supplementary_style_guide/style_guidelines/general-formatting.adoc
```diff
@@ -21,3 +21,72 @@ When the format _day Month year_ causes a presentation or clarity issue, use ins
...
+Product names and abbreviations::
+Use the product name attribute for all instances of the product name where possible.
+Avoid using hard-coded product names.
+
+Product versions::
+Use the product version attribute for instances where the product version should change with the next version and the content would still be correct.
+Do not use the product version attribute if the version should not change.
+For example, if a feature was introduced in a certain version, the version should be hard-coded.
+
+////
```
gregraka, line 81: I realize that it's commented out, but my vote would be to drop the topic **Recommendations on attribute name syntax**. At least for now.

gregraka (approved):
This issue has been around for a while and has been pretty thoroughly reviewed. Good job, @inoxx03 !

supplementary_style_guide/style_guidelines/general-formatting.adoc
gregraka: IMHO, Stefan uses language (i.e. "such as")  that implies that the directories are examples: "...store the file in a common directory above the directories that contain your titles, such as docs/common/ or docs/topics/templates."

supplementary_style_guide/style_guidelines/general-formatting.adoc
inoxx03: thanks:)  that's a good point. I changed _may_ to _can_ or _might_ in several instances throughout the section.

supplementary_style_guide/style_guidelines/general-formatting.adoc
inoxx03: @gregraka I removed the section and incorporated parts of the information from it into the list iunder _Minimum required attributes_

inoxx03:
@gregraka @stoobie thank you for the extensive feedback and reviews on this PR. I revised all the sections based on your input, trimmed the extra text and removed some references that went too far into modular docs convention territory. I also added examples on what the product name and version attributes might look like, making it clear that the naming presented in the examples are not meant to be  prescriptive, and users can choose to follow different naming conventions if they want to.
I also removed the specific recommendation about attribute files and replaced the content of the section with more general recommendation about attribute files. I also added a link to the example mod docs repo by @sterobin (could not find any recommendations on attributes or attribute files in the mod docs guide). Finally, I dropped the recommendations on attribute usage.

Please let me know if the revisions work for you. If there are additional changes that you want me to make, please let me know.

Thank you!:)

supplementary_style_guide/style_guidelines/general-formatting.adoc
gregraka: I see that inoxx03 has removed this content. So, in order of moving this PR along, I'm resolving this conversation.

gregraka (approved):
Hi Stefan (@inoxx03), 
Thanks for all of your work on this! Looks great!
Best,
--Greg

supplementary_style_guide/style_guidelines/general-formatting.adoc
```diff
@@ -100,5 +100,67 @@ Text for note.
...
+----
+:name-product-abbreviated: RHBQ
+----
+
+The major and minor version of the product:: Use an attribute for the product version in cases where the product version can change with each release and the content is still correct.
+For example:
++
+.Example long and short product version attributes
+----
+:version-product-short: 1.11.6
```
inoxx03, line 148:
```suggestion
:version-product-short: 1.11
```

supplementary_style_guide/style_guidelines/general-formatting.adoc
```diff
@@ -100,5 +100,67 @@ Text for note.
...
+:version-product-long: 1.11
```
inoxx03, line 149:
```suggestion
:version-product-long: 1.11.6
```

supplementary_style_guide/style_guidelines/general-formatting.adoc
```diff
@@ -100,5 +100,67 @@ Text for note.
...
+:name-product-abbreviated: RHBQ
+----
+
+The major and minor version of the product:: Use an attribute for the product version in cases where the product version can change with each release and the content is still correct.
+For example:
++
+.Example long and short product version attributes
+----
+:version-product-short: 1.11
+:version-product-long: 1.11.6
```
inoxx03, line 149:
```suggestion
:version-product-long: 1.11.6
:version-product-short: 1.11
```

inoxx03:
@gregraka Thank you for the ACK, I had to make a small correction, because I had reversed the values of the long and the short version attribute examples and have to swap them to correct the example. Sorry about the confusion. Please provide an ACK on the latest changes when you have a chance:) Thank you:)

gregraka (dismissed):
Hi Stefan,
Sorry that I missed that. Glad that you caught it, though!
LGTM!
Thanks,
--Greg

inoxx03:
@redhat-documentation/ccs-style-council I have no further changes to make to this PR. Please let me know whether any additional changes changes are required. If not, please feel free to review and merge. Thank you:)

supplementary_style_guide/style_guidelines/general-formatting.adoc
```diff
@@ -100,5 +100,67 @@ Text for note.
...
+=== Attribute file
+
+Define attributes for product name and product version and store them in a dedicated attributes file for each set of product documentation.
+For examples of where you can store the shared attributes file inside your documentation repository, see the link:https://github.com/redhat-documentation/modular-docs/tree/mod-doc-repo-example[Example modular documentation repository].
+Include the attributes file at the beginning of the `master.adoc` files of all titles in your documentation set:
+
+.master.adoc
+[source,asciidoc,options="nowrap"]
+----
+\include::<path-to-directory-with-attributes-file>/attributes.adoc[]
```
bergerhoffer, line 118: Can you update this replaceable value to follow the SSG format for them? Needs to be italicized and use underscores between the words.

supplementary_style_guide/style_guidelines/general-formatting.adoc
```diff
@@ -100,5 +100,67 @@ Text for note.
...
+[product-names-versions-ref]]
+== Product names and version references
+
+Whenever you refer to the name of your product in full, or in its abbreviated form, or when you refer to the major and minor version of your product, avoid using hard-coded references and use attributes instead.
+Only use hard-coded version references if the version that you are referring to in a particular case never changes.
+
+=== Attribute file
+
+Define attributes for product name and product version and store them in a dedicated attributes file for each set of product documentation.
+For examples of where you can store the shared attributes file inside your documentation repository, see the link:https://github.com/redhat-documentation/modular-docs/tree/mod-doc-repo-example[Example modular documentation repository].
```
bergerhoffer, line 112: From this link, I don't see where the shared attributes file is. Could you update this link (or update the sentence) to give more specifics on where that file is?

supplementary_style_guide/style_guidelines/general-formatting.adoc
```diff
@@ -100,5 +100,67 @@ Text for note.
...
+Include the attributes file at the beginning of the `master.adoc` files of all titles in your documentation set:
+
+.master.adoc
```
bergerhoffer, line 115: Can you update this title to be consistent with our other examples? (with Example in front)

supplementary_style_guide/style_guidelines/general-formatting.adoc
```diff
@@ -100,5 +100,67 @@ Text for note.
...
+----
+:name-product-abbreviated: RHBQ
+----
+
+The major and minor version of the product:: Use an attribute for the product version in cases where the product version can change with each release and the content is still correct.
+For example:
++
+.Example long and short product version attributes
+----
+:version-product-long: 1.11.6
```
bergerhoffer, line 148: Probably nitpicky, but I'd rather see these called minor and major, as opposed to long/short. Long/short makes sense to me in terms of the name, but not necessarily for version numbers.

supplementary_style_guide/style_guidelines/general-formatting.adoc
```diff
@@ -100,5 +100,67 @@ Text for note.
...
+.Example product name attribute
+----
+:name-product: Red Hat build of Quarkus
+----
+
+The abbreviated form of the product name:: If it is necessary for your product, you can use an attribute to store a shortened version of the name of your product, for example:
++
+.Example abbreviated product name attribute
+----
+:name-product-abbreviated: RHBQ
```
bergerhoffer, line 140: Could we use a different product for these examples? I just checked the master naming spreadsheet for this and it specifically says to not use RHBQ as a shortened form.

bergerhoffer (commented):
Just a few minor changes requested, but overall LGTM. It provides some good general guidance while leaving flexibility on what you name them, etc.

supplementary_style_guide/style_guidelines/general-formatting.adoc
inoxx03: OK, but how about _minor_ and _patch_? Technically speaking, _major_ would correspond to the first digit in semantic versioning terminology, and minor would correspond to the second digit. Although it is not  a typical case, some products might need a major version attribute, too, so following the semantic versioning terminology would help avoid potential naming conflicts, WDYT?

inoxx03:
@bergerhoffer Thank you for the review! I've updated the section  to address your feedback. Please let me know what you think:)

supplementary_style_guide/style_guidelines/general-formatting.adoc
```diff
@@ -100,5 +100,67 @@ Text for note.
...
+Whenever you refer to the name of your product in full, or in its abbreviated form, or when you refer to the major and minor version of your product, avoid using hard-coded references and use attributes instead.
+Only use hard-coded version references if the version that you are referring to in a particular case never changes.
+
+=== Attribute file
+
+Define attributes for product name and product version and store them in a dedicated attributes file for each set of product documentation.
+For examples of where you can store the shared attributes file inside your documentation repository, see the link:https://github.com/redhat-documentation/modular-docs/blob/mod-doc-repo-example/_artifacts/document-attributes.adoc[Example modular documentation repository].
+Include the attributes file at the beginning of the `master.adoc` files of all titles in your documentation set:
+
+.Example Asciidoc: Attribute file included in a master.adoc file
```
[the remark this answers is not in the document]
inoxx03: Thank you! Fixed:)

supplementary_style_guide/style_guidelines/general-formatting.adoc
```diff
@@ -100,5 +100,67 @@ Text for note.
...
+.Example Asciidoc: Product name attribute
+----
+:name-product: Red Hat build of Quarkus
+----
+
+The abbreviated form of the product name:: If it is necessary for your product, you can use an attribute to store a shortened version of the name of your product, for example:
++
+.Example Asciidoc: Abbreviated product name attribute
+----
+:name-product-abbreviated: JBoss EAP for OpenShift
```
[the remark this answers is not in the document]
inoxx03: Oh, sorry, I forgot to update the long product name, my bad. Sure, I like your suggestion, and I'll update the examples accordingly. Thank you!

supplementary_style_guide/style_guidelines/general-formatting.adoc
```diff
@@ -100,5 +100,67 @@ Text for note.
...
+----
+
+The major and minor version of the product:: Use an attribute for the product version in cases where the product version can change with each release and the content is still correct.
+For example:
++
+.Example Asciidoc: Long and short product version attributes
```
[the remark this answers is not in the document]
inoxx03: Done!:)

inoxx03:
@bergerhoffer Thank you for the suggestions. I have updated the PR. Please let me know whether the latest revisions are OK.

ndeevy (dismissed):

supplementary_style_guide/style_guidelines/general-formatting.adoc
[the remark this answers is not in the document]
bergerhoffer: @inoxx03 Looks like these product names weren't updated, but everything else looks good

inoxx03:
@bergerhoffer Thasnk! You're right about the product names! I made the update in my editor and didn't save it before pushing -_-
I update the product names now. Apologies about wasting your time.

@ndeevy thank you for the ACK:) Sadly I had messed up the last update and had to push a small fix. Please take another look when you have a chance. Thank you:)

ndeevy (approved):

ndeevy:
Done + approved @inoxx03 LGTM
Thanks a million for taking care of this one :-)

bergerhoffer (approved):
Thanks a lot for the adjustments @inoxx03! This LGTM now. 

With @ndeevy's + @gregraka's approval from earlier, I'm counting that as 3 approvals. Merging!

inoxx03:
Thank you!

PR #30, opened 2020-10-20
Adding guidance for replaceable values

bergerhoffer:
Preview (internal): http://file.rdu.redhat.com/~ahoffer/style_guide/replaceable-values/master.html#user-replaced-values

### DIFF
--- a/supplementary_style_guide/style_guidelines/general-formatting.adoc
+++ b/supplementary_style_guide/style_guidelines/general-formatting.adoc
@@ -10,8 +10,52 @@ Follow _The IBM Style Guide_  guidance of _day Month year_ for date formats, for
 When the format _day Month year_ causes a presentation or clarity issue, use instead: _Month day, year_ (for example, October 3, 2019).
 
 
-[[user-replaced-variables]]
-== User-replaced variables
+[[user-replaced-values]]
+== User-replaced values
+
+A _user-replaced value_ is a value that the user must replace with a value that is relevant for their situation. User-replaced values are often found in places such as code blocks, file paths, and commands.
+
+Use descriptive names for user-replaced values and follow this general format: _<value_name>_.
+
+Ensure that user-replaced values have the following characteristics:
+
+* Surrounded by angle brackets (`< >`)
+* Separated by underscores (`_`) for multi-word values
+* Lowercase, unless the rest of the related text is uppercase or another capitalization scheme.
+* Italicized
+* If the user-replaced value is referencing a value in code or in a command that is normally monospace, also use monospace for the user-replaced value.
+
+.Example AsciiDoc: User-replaced value in a paragraph
+
+----
+Create an Ansible inventory file that is named `/_<path>_/inventory/hosts`.
+----
+
+This renders as follows:
+
+====
+Create an Ansible inventory file that is named `/_<path>_/inventory/hosts`.
+====
+
+To italicize a user-replaced value in a code block, you must add an attribute to apply text formatting, such as `subs="+quotes"` or `subs="normal"`, to the attribute list of the code block.
+
+.Example AsciiDoc: User-replaced value in a code block
+
+  [subs="+quotes"]
+  ----
+  $ oc describe node __<node_name>__
+  ----
+
+This renders as follows:
+
+====
+[subs="+quotes"]
+----
+$ oc describe node __<node_name>__
+----
+====
+
+// TODO: In the future, add alternative guidance for XML
 
 [[lead-in-sentences]]
 == Lead-in sentences


bergerhoffer:
@redhat-documentation/ccs-style-council Can we get some reviews on this please?

supplementary_style_guide/style_guidelines/general-formatting.adoc
```diff
@@ -10,8 +10,52 @@ Follow _The IBM Style Guide_  guidance of _day Month year_ for date formats, for
 When the format _day Month year_ causes a presentation or clarity issue, use instead: _Month day, year_ (for example, October 3, 2019).
 
 
-[[user-replaced-variables]]
-== User-replaced variables
+[[user-replaced-values]]
+== User-replaced values
+
+A _user-replaced value_, or _replaceable value_, is a value that must be replaced by the user with a value that is relevant for their situation. User-replaced values are often found in places such as code blocks, file paths, and commands.
```
ndeevy, line 16:
```suggestion
A _user-replaced value_, or _replaceable value_, is a value that the user must replace with a value that is relevant for their situation. User-replaced values are often found in places such as code blocks, file paths, and commands.
```
Suggestion to make passive construction, active.

supplementary_style_guide/style_guidelines/general-formatting.adoc
```diff
@@ -10,8 +10,52 @@ Follow _The IBM Style Guide_  guidance of _day Month year_ for date formats, for
...
+
+User-replaced values should use a descriptive names, and should follow this general format: _<value_name>_.
```
ndeevy, line 18:
```suggestion
Use descriptive names for user-replaced values and follow this general format: _<value_name>_.
```

Suggestion to make the sentence more user-action oriented and prescriptive and replace the 'should' so it's more direct.

supplementary_style_guide/style_guidelines/general-formatting.adoc
```diff
@@ -10,8 +10,52 @@ Follow _The IBM Style Guide_  guidance of _day Month year_ for date formats, for
...
+
+Replaceable values should have the following characteristics:
```
ndeevy, line 20: To replace the 'should' to be more direct per IBM, you could change this to something like one of these options:

- Use the following format for user-replaced values:
- Replaceable values must have the following characteristics:
- Ensure that user-replaced values have following characteristics:

It might also be nicer to keep referring to them as 'user-replaced values' for consistency throughout the body of the entry?

supplementary_style_guide/style_guidelines/general-formatting.adoc
```diff
@@ -10,8 +10,52 @@ Follow _The IBM Style Guide_  guidance of _day Month year_ for date formats, for
...
+
+* Surrounded by angle brackets (`< >`)
+* Separated by underscores (`_`) for multi-word values
+* Lowercase, unless the rest of the related text is uppercase or another capitalization scheme.
+* Italicized
+* If the replaceable value is referencing a value in code or in a command that would normally be monospace, then the reference to the replaceable value should also be monospace.
```
ndeevy, line 26:
```suggestion
* If the replaceable value is referencing a value in code or in a command that is normally monospace, use monospace for the replaceable value also.
```

Suggestion to be more direct and prescriptive.

supplementary_style_guide/style_guidelines/general-formatting.adoc
```diff
@@ -10,8 +10,52 @@ Follow _The IBM Style Guide_  guidance of _day Month year_ for date formats, for
...
+Create an Ansible inventory file that is named `/_<path>_/inventory/hosts`.
+----
+
+This renders as follows:
+
+====
+Create an Ansible inventory file that is named `/_<path>_/inventory/hosts`.
+====
+
+To italicize a replaceable value in a code block, you must add `subs="+quotes"` to the code block's attribute list.
```
ndeevy, line 40:
```suggestion
To italicize a replaceable value in a code block, you must add `subs="+quotes"` to the attribute list of the code block.
```

Suggestion to be more fully in line with IBM style.

ndeevy:
Did a quick review @bergerhoffer Looks good and it's great to have cross-product guidance on this!

bergerhoffer:
Thanks for the great feedback @ndeevy! I've updated to address your comments.

supplementary_style_guide/style_guidelines/general-formatting.adoc
```diff
@@ -10,8 +10,52 @@ Follow _The IBM Style Guide_  guidance of _day Month year_ for date formats, for
...
-[[user-replaced-variables]]
-== User-replaced variables
+[[user-replaced-values]]
+== User-replaced values
+
+A _user-replaced value_, or _replaceable value_, is a value that the user must replace with a value that is relevant for their situation. User-replaced values are often found in places such as code blocks, file paths, and commands.
+
+Use descriptive names for user-replaced values and follow this general format: _<value_name>_.
+
+Ensure that user-replaced values have following characteristics:
```
ndeevy, line 20:
```suggestion
Ensure that user-replaced values have the following characteristics:
```

One more tiny suggestion to add 'the' back in.

bergerhoffer: Whoops, good catch sorry!

bergerhoffer:
@ndeevy or @redhat-documentation/ccs-style-council Can you take one last look through this and if there is no additional feedback, can you approve it? Thanks!

supplementary_style_guide/style_guidelines/general-formatting.adoc
```diff
@@ -10,8 +10,52 @@ Follow _The IBM Style Guide_  guidance of _day Month year_ for date formats, for
 When the format _day Month year_ causes a presentation or clarity issue, use instead: _Month day, year_ (for example, October 3, 2019).
 
 
-[[user-replaced-variables]]
-== User-replaced variables
+[[user-replaced-values]]
+== User-replaced values
+
+A _user-replaced value_, or _replaceable value_, is a value that the user must replace with a value that is relevant for their situation. User-replaced values are often found in places such as code blocks, file paths, and commands.
```
stoobie, line 16: @bergerhoffer 
1. Generally only use one term, unless there's a good reason to use more than one. Since you're defining the terms here, is there any reason to include both terms _user-replaced value_ and _replaceable value_?
2. Please add a short example. Suggestion:
```suggestion
A _user-replaced value_  is a value that the user must replace. For example, in the following command `_filename_` and `_destination_` are user-replaced values.

`cp _filename_ _destination_`

User-replaced values are often found in places such as code blocks, file paths, and commands.
```

supplementary_style_guide/style_guidelines/general-formatting.adoc
```diff
@@ -10,8 +10,52 @@ Follow _The IBM Style Guide_  guidance of _day Month year_ for date formats, for
...
+Create an Ansible inventory file that is named `/_<path>_/inventory/hosts`.
+----
+
+This renders as follows:
+
+====
+Create an Ansible inventory file that is named `/_<path>_/inventory/hosts`.
+====
+
+To italicize a user-replaced value in a code block, you must add `subs="+quotes"` to the attribute list of the code block.
```
stoobie, line 40: @bergerhoffer I'm not an expert on the code block macros, but are you positive this is a requirement? Here's an example from our docs:
```
[options="nowrap" subs="normal"]
----
Host fully qualified DNS name of this server [_autodetected host name_]:
----
```
And I _think_ that when you have additional underscores in the variable, you should (must?) use double-underscores for italicizing, e.g.: __multi_word_variable__

If I'm wrong, I'd love to learn more.

stoobie (changes requested):

supplementary_style_guide/style_guidelines/general-formatting.adoc
bergerhoffer: @stoobie In order for the italics to render, you need to add something to the attribute list to tell it to do so. `quotes` is the attribute that performs text styling substitutions. It looks like your example uses the `normal` attribute, which looks like it adds text styling and all other substitutions except callouts [1]. So yes the `normal` one seems to work fine too, but I'm not sure if we'd want to use that one as the suggestion, as it performs additional substitutions that might not be wanted. Let me know if you do think we should mention it as an alternative though.

[1] https://asciidoctor.org/docs/asciidoc-syntax-quick-reference/#attributes-and-substitutions

As for the double underscores, you definitely don't have to use them if the word has underscores, but you are right that there are some situations where the double underscores might be necessary. I can't remember exactly what those are - I think it might be when there is some additional formatting. However, I don't see any harm in using the double underscores in our example. I'll update.

supplementary_style_guide/style_guidelines/general-formatting.adoc
bergerhoffer: @stoobie I updated to remove the second term mention of "replaceable value". 

The examples are further down after we explain the guidelines. Are those not sufficient?

bergerhoffer:
@stoobie Can you take another look through after my updates, thanks!

supplementary_style_guide/style_guidelines/general-formatting.adoc
stoobie: @bergerhoffer It sounds like you have a better grasp on the code block macros than I do. However, from what you wrote, it's not accurate to say that you must add `subs="+quotes"` if there are other alternatives. I think it would be nice to say that there are alternatives, and include a link to the quick reference you included here.

bergerhoffer: @stoobie That page doesn't have a great anchor for that attribute section, so I don't necessarily want to link to that. But I've updated to include both quotes and normal as examples of attributes that apply text formatting. Let me know what you think!

stoobie (approved):
LGTM

PR #52, opened 2020-12-15
Fixing line that was incorrectly combined with the previous paragraph

bergerhoffer:
Added line break so that the "Use it" line wasn't combined with the previous paragraph.

Also fixed she -> they in an example, to adhere with the "she" entry which recommends to avoid she/he.


cbudz (approved):
LGTM @bergerhoffer

PR #66, opened 2021-02-25
revised f.adoc to add an before adjective in two places

leswilliams44:
Added "an" before adjective in two places


bergerhoffer (approved):

bergerhoffer:
Merging, as this is just a minor grammar update.

PR #71, opened 2021-03-10
added entry for Conversational style

mportman12:
Add a section to the General Formatting section for conversational style guidance.

### DIFF
--- a/supplementary_style_guide/style_guidelines/general-formatting.adoc
+++ b/supplementary_style_guide/style_guidelines/general-formatting.adoc
@@ -17,11 +17,9 @@ Follow _The IBM Style Guide_ guidance of _less conversational_ style in most cas
 
 As needed, adjust the conversational to _fairly conversational_ for an audience of new users or _least conversational_ for API documentation and other very experienced audiences.
 
-For example:
+.Example: Less conversational style
 
-====
 Red Hat Enterprise Linux 8 delivers a stable, secure, and consistent foundation across hybrid cloud deployments with the tools needed to deliver workloads faster with less effort.
-====
 
 [[user-replaced-values]]
 == User-replaced values
--- a/supplementary_style_guide/style_guidelines/general-formatting.adoc
+++ b/supplementary_style_guide/style_guidelines/general-formatting.adoc
@@ -17,12 +17,11 @@ Follow _The IBM Style Guide_ guidance of _less conversational_ style in most cas
 
 As needed, adjust the conversational to _fairly conversational_ for an audience of new users or _least conversational_ for API documentation and other very experienced audiences.
 
-.Example AsciiDoc: Less conversational style 
+For example:
 
-----
-Get answers now. Search the documentation and posts from the Stack Overflow community, which is actively monitored.
-The jobs in a stage cannot pass artifacts to each other.
-----
+====
+Red Hat Enterprise Linux 8 delivers a stable, secure, and consistent foundation across hybrid cloud deployments with the tools needed to deliver workloads faster with less effort.
+====
 
 [[user-replaced-values]]
 == User-replaced values
--- a/supplementary_style_guide/style_guidelines/general-formatting.adoc
+++ b/supplementary_style_guide/style_guidelines/general-formatting.adoc
@@ -10,6 +10,20 @@ Follow _The IBM Style Guide_  guidance of _day Month year_ for date formats, for
 When the format _day Month year_ causes a presentation or clarity issue, use _Month day, year_ (for example, October 3, 2019) instead.
 
 
+[[conversational-style]]
+== Conversational style
+
+Follow _The IBM Style Guide_ guidance of _less conversational_ style in most cases. 
+
+As needed, adjust the conversational to _fairly conversational_ for an audience of new users or _least conversational_ for API documentation and other very experienced audiences.
+
+.Example AsciiDoc: Less conversational style 
+
+----
+Get answers now. Search the documentation and posts from the Stack Overflow community, which is actively monitored.
+The jobs in a stage can't pass artifacts to each other.
+----
+
 [[user-replaced-values]]
 == User-replaced values
 
--- a/supplementary_style_guide/style_guidelines/general-formatting.adoc
+++ b/supplementary_style_guide/style_guidelines/general-formatting.adoc
@@ -21,7 +21,7 @@ As needed, adjust the conversational to _fairly conversational_ for an audience 
 
 ----
 Get answers now. Search the documentation and posts from the Stack Overflow community, which is actively monitored.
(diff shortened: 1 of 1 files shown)


mportman12:
@redhat-documentation/ccs-style-council For review please

leswilliams44:
There is a new group that meets to define guidelines for managed services. Conversational tone was a large part of today's meeting. You can see the agenda and notes here: 

https://docs.google.com/document/d/1UTyf9Aq5GwcSrqSmUrqtDDWKFyWnOkjuMvoPLPxv1XE/edit#heading=h.psgow3esbv7u

It might be worthwhile to talk with @cbudz because today's meeting covered a lot of scenarios for what tone to use.

mportman12:
@cbudz Thoughts? 
I think it's only worthwhile to define the Red Hat conversational style if there's one style that fits the majority of product documentation. If Managed Services can envision scenarios for all four IBM conversational styles within our product documentation, there's no point adding a recommendation to the supplementary style guide.

cbudz:
hey @mportman12 - this initial discussion lead to potentially assigning different levels of tone between concept and procedure module types for Managed Services. Adopting the 'fairly conversational" tone for concepts and the "less conversational" tone for procedures since this sort of work will be used in Quick Starts, tours, etc.

bergerhoffer:
@cbudz Do you see any reasons why we shouldn't call out a default conversational style for product docs? I can understand adapting the style as appropriate, like it sounds like managed services will be doing if the content is being reused for quick starts and tutorials. But I'd think it could still be appropriate to define our normal level as less conversational, and as the PR does here, says when it might be appropriate to follow other tones. wdyt?

ndeevy:
+1  to define our normal level as **Less conversational** (or even **Least conversational**) because as tech docs go, I think Red Hat's are for a pretty technical/expert audience.

For consistency I'd advocate for keeping the same tone for both concept and procedure modules.

Another thing to consider about moving more towards the more conversational level is international audiences. The **Fairly conversational** level example includes 'dive in', which is an idiomatic phrasal verb, which doesn't help translation or global audiences. And "Get answers now." seems more appropriate for marketing material. The word "get" has tons of meanings when translated and is often used in phrasal verbs so even in quickstarts etc. I still think we should err on the side of accuracy by using words that have less chance of being misinterpreted. Being conversational is good but being understood feels more important.

In any case, I'd vote to exclude these particular less conversational examples ^^ in our guide.

stoobie:
+1 To what @ndeevy said.

ndeevy:
Thanks for this @mportman12 
For consistency, non-native readers (our majority) and translation, I still would prefer: use 'least formal' or  'less formal' in general unless your product area prefers 'fairly conversational' for new users.
But, I'm happy to go with the team's decision on this :-]

supplementary_style_guide/style_guidelines/general-formatting.adoc
```diff
@@ -10,6 +10,20 @@ Follow _The IBM Style Guide_  guidance of _day Month year_ for date formats, for
...
+
+Follow _The IBM Style Guide_ guidance of _less conversational_ style in most cases. 
+
+As needed, adjust the conversational to _fairly conversational_ for an audience of new users or _least conversational_ for API documentation and other very experienced audiences.
+
+.Example AsciiDoc: Less conversational style 
+
+----
+Get answers now. Search the documentation and posts from the Stack Overflow community, which is actively monitored.
+The jobs in a stage can't pass artifacts to each other.
```
bergerhoffer, line 24: I'd suggest against using a contraction in this example. I know that the new ISG allows it, but I'll open an issue for discussion on whether we should add guidance to continue avoiding them or not.

I'm also unsure whether we want to use exact examples from the ISG or if we should come up with our own.

ndeevy: +1 
Our own examples could avoid things that might be problematic for translation too, like colloquialisms or phrasal verbs.

mportman12:
Removed the problem contraction. I found a few suggestions to replace the ISG example:

a. "Red Hat Enterprise Linux 8 delivers a stable, secure, consistent foundation across hybrid cloud deployments with the tools needed to deliver workloads faster with less effort." 
- from Chapter 1 of Performing a standard RHEL installation

b. "To provide services for large numbers of clients, as well as for redundancy and availability, IdM allows deployment on multiple IdM servers in a single domain. It is possible to deploy up to 60 servers." 
- from Chapter 1 of Planning Identity Management

c. "Running the container tools such as podman, skopeo, or buildah as a user with superuser privilege (root user) is the best way to ensure that your containers have full access to any feature available on your system." 
- from Chapter 1 of Building, running, and managing containers

Does anyone have a preference?

leswilliams44:
I like option c because it is more direct.

leswilliams44 (approved):
LGTM

leswilliams44 (dismissed):
LGTM

supplementary_style_guide/style_guidelines/general-formatting.adoc
```diff
@@ -10,6 +10,20 @@ Follow _The IBM Style Guide_  guidance of _day Month year_ for date formats, for
...
+== Conversational style
+
+Follow _The IBM Style Guide_ guidance of _less conversational_ style in most cases. 
+
+As needed, adjust the conversational to _fairly conversational_ for an audience of new users or _least conversational_ for API documentation and other very experienced audiences.
+
+.Example AsciiDoc: Less conversational style 
+
+----
+Get answers now. Search the documentation and posts from the Stack Overflow community, which is actively monitored.
```
stoobie, line 23: This isn't really an Asciidoc example, it's a writing example. I would remove the Asciidoc markup and remove Asciidoc from the heading.

We provide Asciidoc examples to demonstrate how to code a specific guideline. That guidance is moot in this example.

(BTW, is it AsciiDoc or Asciidoc?)

stoobie (changes requested):

supplementary_style_guide/style_guidelines/general-formatting.adoc
bergerhoffer: +1 for removing "AsciiDoc" from the example title, can just be "Example: Less conversational style". And the code block doesn't seem necessary. Could do it like how the linking section does it: https://redhat-documentation.github.io/supplementary-style-guide/#rh-kb-links

And yeah it is AsciiDoc capital D.

mportman12: According to the asciidoc.org, it's AsciiDoc. Remove AsciiDoc markup.

leswilliams44 (approved):
LGTM
Thanks for adding "and."

bergerhoffer:
This has enough approvals, merging.

PR #72, opened 2021-03-11
AMD64: Clarify that it is 64-bit x86, and not 32-bit, etc. x86

wking:
E.g. [Wikipedia][1]:

> x86-64 (also known as x64, x86_64, AMD64 and Intel 64)[note 1] is a 64-bit version of the x86 instruction set...

and [also][2]:

> Introduced  1978 (16-bit), 1985 (32-bit), 2003 (64-bit)

Also "a ... version" and not "the ... version", because there are presumably other ways you could extend x86 to 64 bits.

[1]: https://en.wikipedia.org/wiki/X86-64
[2]: https://en.wikipedia.org/wiki/X86

### DIFF
--- a/supplementary_style_guide/glossary_terms_conventions/general_conventions/a.adoc
+++ b/supplementary_style_guide/glossary_terms_conventions/general_conventions/a.adoc
@@ -46,7 +46,7 @@
 [discrete]
 [[AMD64]]
 ==== AMD64 (noun)
-*Description*: "AMD64" is the AMD implementation of the x86 architecture.
+*Description*: "AMD64" is the AMD implementation of a 64-bit version of the x86 architecture.
 
 *Use it*: yes
 


bergerhoffer (approved):
Thanks @wking for submitting this! This LGTM

@redhat-documentation/ccs-style-council for review as well

cbudz:
LGTM 👍

stoobie (changes requested):
This is not exactly correct and complete. AMD64 is a non-formal nickname used to refer to the 64-bit version of the x86 architecture. AMD initiated a 64-bit implementation of x86, hence the nickname. However, Intel also has their own 64-bit implementation of the x86 architecture. Intel refers to this architecture as Intel(R) 64. A generic term is x86-64.

See https://en.wikipedia.org/wiki/X86-64

stoobie:
I see now that @wking referenced that same Wikipedia article. I think that the right thing for us to do is _not_ to show preference to one implementation vs another in our style guide.

bergerhoffer:
> I see now that @wking referenced that same Wikipedia article. I think that the right thing for us to do is _not_ to show preference to one implementation vs another in our style guide.

@stoobie I'm not sure what you mean. It's just a term definition, I'm not sure that that's showing preference?

stoobie:
@bergerhoffer The definition of the term AMD64 is not in itself showing preference. The fact that our glossary only includes a definition for AMD64 and not for Intel64 or the generic x86-64 doesn't take into account that AMD is not the only vendor to provide a 64-bit x86-compatible architecture.

The correct thing to do is to define x86-64 and to provide guidance to _not_ use AMD64 or Intel64. I realize that does not reflect the reality that a lot of docs use AMD64, but as far as guidance, it doesn't make sense to use a non-generic term when a generic term exists. I don't believe that AMD64 is more recognizable to x86-64.

bergerhoffer:
@stoobie I'm not sure whether it'd be right to recommend not to use AMD64 or Intel64, because there might be places where that is appropriate, but I think it's at least fair to add entries for the others, if you wanted to do that.

gregraka:
Here's one use case of these terms in Red Hat OpenStack Platform documentation:

"An 8-core 64-bit x86 processor with support for the Intel 64 or AMD64 CPU extensions."

From: https://access.redhat.com/documentation/en-us/red_hat_openstack_platform/13/html-single/director_installation_and_usage/index#sect-Undercloud_Requirements

stoobie:
I think mentioning both Intel64 and AMD64 is fine, @gregraka. However, I want to point out that by definition, an x86 processor is a 32-bit processor. I would rewrite the sentence like this:
"An 8-core 64-bit processor with support for the Intel64 or AMD64 CPU extensions." or
"An 8-core 64-bit x86-64 processor with support for the Intel64 or AMD64 CPU extensions."

Although I think the second sentence is redundant.

gregraka:
Thanks, Steve (@stoobie). I will pass this along to the writer.

PR #73, opened 2021-03-11
Adding a history page for updates

bergerhoffer:
Adding a page where we list the significant updates to the guide. Current use case for it is to link to from email updates, for people to see what's new.


cbudz (approved):
LGTM

PR #77, opened 2021-05-03
Add alts for "user-replaced value"

adellape:
* I've seen some questions recently about combining the `subs` attr when also applying source highlighting, so I've added an example showing the comma-style syntax for listing multiple.
* Threw in some AKAs for "user-replaced values", since it might be helpful for people trying to CTRL+F the guide (many CCS folks refer to them still as "replaceables" and the IBM SG calls them "variable values").

~~Added `subs="attributes+"` to the list of suggestions, as it's the one used the most in the `openshift-docs` repo.~~

### DIFF
--- a/supplementary_style_guide/style_guidelines/general-formatting.adoc
+++ b/supplementary_style_guide/style_guidelines/general-formatting.adoc
@@ -13,7 +13,7 @@ When the format _day Month year_ causes a presentation or clarity issue, use _Mo
 [[user-replaced-values]]
 == User-replaced values
 
-A _user-replaced value_ is a value that the user must replace with a value that is relevant for their situation. User-replaced values are often found in places such as code blocks, file paths, and commands.
+A _user-replaced value_, also known as a replaceable or variable value, is a value that the user must replace with a value that is relevant for their situation. User-replaced values are often found in places such as code blocks, file paths, and commands.
 
 Use descriptive names for user-replaced values and follow this general format: _<value_name>_.
 


supplementary_style_guide/style_guidelines/general-formatting.adoc
```diff
@@ -37,7 +37,7 @@ This renders as:
 Create an Ansible inventory file that is named `/_<path>_/inventory/hosts`.
 ====
 
-To italicize a user-replaced value in a code block, you must add an attribute to apply text formatting, such as `subs="+quotes"` or `subs="normal"`, to the attribute list of the code block.
+To italicize a user-replaced value in a code block, you must add a substitution attribute to apply text formatting to the attribute list of the code block. You can use `subs="+quotes"`,  `subs="normal"`, or `subs="attributes+"`.
```
bergerhoffer, line 40: @adellape I thought that `subs="attributes+"` was for if you're replacing an attribute in a code block. Like {product-title} and wanting it to resolve to "OpenShift Container Platform". Unless I'm mistaken, I don't think that you can use `subs="attributes+"` to italicize in a code block.

supplementary_style_guide/style_guidelines/general-formatting.adoc
```diff
@@ -13,7 +13,7 @@ When the format _day Month year_ causes a presentation or clarity issue, use _Mo
 [[user-replaced-values]]
 == User-replaced values
 
-A _user-replaced value_ is a value that the user must replace with a value that is relevant for their situation. User-replaced values are often found in places such as code blocks, file paths, and commands.
+A _user-replaced value_, also known as a replaceable or variable value, is a value that the user must replace with a value that is relevant for their situation. User-replaced values are often found in places such as code blocks, file paths, and commands.
```
bergerhoffer, line 16: +1 for these AKAs

bergerhoffer (commented):
I like the AKAs, but am uncertain about the accuracy of including attributes+ as a valid option to italicize text, and the other AsciiDoc-specific items.

supplementary_style_guide/style_guidelines/general-formatting.adoc
```diff
@@ -55,6 +55,16 @@ $ oc describe node __<node_name>__
 ----
 ====
 
+When applying multiple attributes to a block, use the following syntax:
```
bergerhoffer, line 58: I meant to leave a comment here but apparently forgot to save it. I think that this would probably be more appropriate to be added to the [AsciiDoc guide](https://redhat-documentation.github.io/asciidoc-markup-conventions/). wdyt?

supplementary_style_guide/style_guidelines/general-formatting.adoc
adellape: Sorry you're right, I started conflating too many things when looking up examples using multiple styles per block. I'll rewind the `attributes+`-related changes.

Though I do notice now that neither the Supplementary Guide or the [openshift-docs guidelines](https://github.com/openshift/openshift-docs/blob/master/contributing_to_docs/doc_guidelines.adoc) mention `attributes+` for the attribute-replacement use case. I imagine other CCS teams might be using attributes in a similar way, but I'm not sure... If they do, maybe worth a follow-up PR for the SG.

supplementary_style_guide/style_guidelines/general-formatting.adoc
adellape: Maybe? I've trimmed down the example so it's not so beefy now, and it would be nice for the tip to be in a place where it's probably pretty likely to come up (not to mention there's already an example syntax right above it. Are you suggesting another entry added to the big table specifically about combining attributes?

bergerhoffer: I'm on the fence about this. Maybe if it was wrapped as a TIP admonition?

I'll defer to others on the @redhat-documentation/ccs-style-council if they think it's okay to add this tip here.

ndeevy: If it's a general tip rather than a directive maybe the AsciiDoc guide is the best place ... keeping the focus here on directives would be more in line with minimalism principles. I don't think we currently have lots of TIPs included.

It's a small one though so wouldn't create a big problem to have it here with a TIP admonition either.

adellape: @bergerhoffer @ndeevy I've updated it as a TIP admonition block.

ndeevy (dismissed):

supplementary_style_guide/style_guidelines/general-formatting.adoc
stoobie: @adellape @bergerhoffer I'm confused about what you're trying to say here...is the tip to use `[source,terminal,subs="+quotes"]` for code blocks that include user-replaceable values? If yes, then I think this will add more confusion...or maybe add a link to detailed info on how to use these attributes or whatever you call them for code blocks. I'm not sure about other writers, but I don't have a real great grasp on all the different options available for code blocks...I tend to reuse what I already see for a similar usage in the doc I'm working on.

stoobie (dismissed):

mportman12:
@bergerhoffer @adellape To keep things moving along, can you address @stoobie's requested changes?

inoxx03:
@bergerhoffer @adellape To keep things moving along, can you address @stoobie's requested changes and rebase this PR to resolve the merge conflicts when you have a chance. Thank you:)

adellape:
@inoxx03 Dropped the `[TIP]` so it's just the AKA now.

ndeevy (approved):

inoxx03 (approved):

PR #80, opened 2021-05-06
Added info about single step procedures

mportman12:
Added guidance about using an unnumbered bullet for single step procedures.


bergerhoffer (approved):
LGTM!

ndeevy (approved):

leswilliams44 (commented):
Should "Single step procedures" be "Single-step procedures" if single step is a compound modifier?

mportman12:
> Should "Single step procedures" be "Single-step procedures" if single step is a compound modifier?

@leswilliams44 Added hyphen

bergerhoffer:
Merging (enough approvals when including those from prior to the last tweak).

PR #84, opened 2021-05-25
Reordered glossary entry table so <class> was in the right place

ctauchen:
I moved the table around so <class> was is the same position as in the template: between <item> and <description>.

Easier to read, I think.


bergerhoffer (approved):
This makes sense to me, merging. Thanks @ctauchen!

PR #85, opened 2021-05-26
Replaced "Internet" with "internet" according to ISG guidelines

ctauchen:
[Revised 2021-06-23]

Following the guidance of ISG, I replaced the capitalized "Internet" with the lowercased "internet" throughout the guide.

[original PR text]
~~I'm adding a glossary entry for "Internet" because:~~

* ~~we tend to use the capitalized form in our docs~~
* ~~without this entry, writers must follow IBM, which prefers the lowercase "internet"~~

~~For the record, I prefer the lowercase form. Style guides and dictionaries are all moving in this direction, with the notable exception of Merriam Webster.~~

~~But until we're prepared to make that commitment, we need to provide clear guidance to writers.~~

~~I've also changed a few instances of "internet" to "Internet" in this guide.~~

### DIFF
--- a/supplementary_style_guide/glossary_terms_conventions/general_conventions/i.adoc
+++ b/supplementary_style_guide/glossary_terms_conventions/general_conventions/i.adoc
@@ -167,7 +167,7 @@
 [discrete]
 [[ip-masquerade]]
 ==== IP Masquerade (noun)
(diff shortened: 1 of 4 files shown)


bergerhoffer:
I've often seen "Internet" capitalized in the docs because that's how the old IBM style guide said to do it. The new one does say to use lowercase:

image

Though elsewhere in the guide it still uses capital, which is a bit confusing.

However, unless we have good reason to deviate, I'd vote for keeping with IBM's new guidance and moving to lowercase "internet".

bergerhoffer:
@ctauchen We discussed this at the 23 June style council meeting, and we agreed to follow IBM's guidance of lowercase "internet". 

Would you mind updating this PR to remove the new entry you added, and instead update all instances of "Internet" to "internet" in this repo?

I'll also update our [history](https://github.com/redhat-documentation/supplementary-style-guide/blob/master/HISTORY.md) file to indicate this decision, even though it didn't result in a new entry.

ctauchen:
Updated PR to remove glossary entry "Internet", replaced instances of "Internet" with "internet."
@bergerhoffer

bergerhoffer (approved):
Thanks a lot for the update @ctauchen! This LGTM, merging

PR #113, opened 2021-08-25
Update Operator conventions

sterobin:
Per discussion in today's style meeting, I've made the following changes for the term "Operator":

* Duplicated the "Operator" entry to the general conventions instead of just the OpenShift conventions.
* Updated the "Operator" entry with some helpful info from the [OpenShift doc guidelines](https://github.com/openshift/openshift-docs/blob/main/contributing_to_docs/doc_guidelines.adoc#operator-capitalization) on the same entry.
* Expanded the entry to clarify capitalization considerations and examples, and to clarify that other types of Kubernetes components that remain lowercase when general or monospaced and proper when specific.
* Updated the same entry in the OpenShift conventions page for consistency between the two.

Entry previews:
* [Operator](https://github.com/sterobin/supplementary-style-guide/blob/operator-convention/supplementary_style_guide/glossary_terms_conventions/general_conventions/o.adoc#-operator-noun) entry in general conventions
* [Operator](https://github.com/sterobin/supplementary-style-guide/blob/operator-convention/supplementary_style_guide/glossary_terms_conventions/product_conventions/openshift.adoc#operator) entry in OpenShift conventions (now the same)

### DIFF
--- a/supplementary_style_guide/glossary_terms_conventions/general_conventions/o.adoc
+++ b/supplementary_style_guide/glossary_terms_conventions/general_conventions/o.adoc
@@ -174,12 +174,14 @@ using the Kubernetes APIs and `kubectl` or `oc` tooling.
 
 The term "Operator" in the context of Kubernetes is always capitalized to distinguish it from other types of operators, such as human or mathematical operators. For example:
 
+.Example: Kubernetes Operator
 ----
 = Support policy for unmanaged Operators
 
 Individual Operators have a `managementState` parameter in their configuration.
 ----
 
+.Example: Mathematical operator
 ----
 The following operators and operands are supported in Drools Rule Language:
 
@@ -207,7 +209,7 @@ NOTE: When referring generally to other Kubernetes components, such as pods, nod
 
 *Incorrect forms*: operator, operatorize
 
-*See also*: link:https://github.com/openshift/openshift-docs/blob/main/contributing_to_docs/doc_guidelines.adoc#api-object-formatting[API object formatting] in OpenShift documentation guidelines
+*See also*: xref:api-objects[API objects]
 
 [discrete]
 [[organization-administrator]]
--- a/supplementary_style_guide/glossary_terms_conventions/general_conventions/o.adoc
+++ b/supplementary_style_guide/glossary_terms_conventions/general_conventions/o.adoc
@@ -207,7 +207,7 @@ NOTE: When referring generally to other Kubernetes components, such as pods, nod
 
 *Use it*: yes
 
-*Incorrect forms*: operator, operatorize
+*Incorrect forms*: Kubernetes operator, operatorize
 
 *See also*: xref:api-objects[API objects]
 
--- a/supplementary_style_guide/glossary_terms_conventions/general_conventions/o.adoc
+++ b/supplementary_style_guide/glossary_terms_conventions/general_conventions/o.adoc
@@ -168,11 +168,9 @@
 [[operator]]
 ==== image:images/yes.png[yes] Operator (noun)
 *Description*: In the context of Kubernetes, an Operator is a method of packaging, deploying, and managing a
-Kubernetes application. A Kubernetes application is an application that is both
-deployed on a Kubernetes cluster (including OpenShift clusters) and managed
-using the Kubernetes APIs and `kubectl` or `oc` tooling.
+Kubernetes application. A Kubernetes application is an application that is both deployed on a Kubernetes cluster (including OpenShift clusters) and managed using the Kubernetes APIs and `kubectl` or `oc` tooling.
 
-The term "Operator" in the context of Kubernetes is always capitalized to distinguish it from other types of operators, such as human or mathematical operators. For example:
+The term "Operator" in the context of Kubernetes is always capitalized to distinguish it from other types of operators, such as human or mathematical operators.
 
 .Example: Kubernetes Operator
 ----
@@ -191,7 +189,7 @@ The following operators and operands are supported in Drools Rule Language:
 ----
 
 An Operator's full name must be a proper noun, with each word initially
-capitalized. If it includes a product name, defer the product's capitalization
+capitalized. If it includes a product name, defer to the product's capitalization
 style guidelines. For example:
 
 - Red Hat OpenShift Logging Operator
@@ -200,7 +198,7 @@ style guidelines. For example:
 - Node Tuning Operator
 - Cluster Version Operator
 
-While "containerized" is allowed, do not use "Operatorize" to refer to building
+Although "containerized" is allowed, do not use "Operatorize" to refer to building
 an Operator that packages an application.
 
 NOTE: When referring generally to other Kubernetes components, such as pods, nodes, or image streams, use lowercase. When referring to a specific component, follow the capitalization of the component name and apply monospace formatting, such as "the `Pod` spec", "a `Node` object", or "an `ImageStream` resource".
--- a/supplementary_style_guide/glossary_terms_conventions/general_conventions/o.adoc
+++ b/supplementary_style_guide/glossary_terms_conventions/general_conventions/o.adoc
@@ -165,6 +165,51 @@
 *See also*:
 
 [discrete]
+[[operator]]
+==== image:images/yes.png[yes] Operator (noun)
+*Description*: In the context of Kubernetes, an Operator is a method of packaging, deploying, and managing a
+Kubernetes application. A Kubernetes application is an application that is both
+deployed on a Kubernetes cluster (including OpenShift clusters) and managed
+using the Kubernetes APIs and `kubectl` or `oc` tooling.
+
+The term "Operator" in the context of Kubernetes is always capitalized to distinguish it from other types of operators, such as human or mathematical operators. For example:
+
+----
+= Support policy for unmanaged Operators
+
+Individual Operators have a `managementState` parameter in their configuration.
+----
+
+----
+The following operators and operands are supported in Drools Rule Language:
+
+* + (addition)
+* - (subtraction)
+...
+----
+
+An Operator's full name must be a proper noun, with each word initially
+capitalized. If it includes a product name, defer the product's capitalization
+style guidelines. For example:
+
+- Red Hat OpenShift Logging Operator
+- Prometheus Operator
+- etcd Operator
+- Node Tuning Operator
+- Cluster Version Operator
+
+While "containerized" is allowed, do not use "Operatorize" to refer to building
+an Operator that packages an application.
+
+NOTE: When referring generally to other Kubernetes components, such as pods, nodes, or image streams, use lowercase. When referring to a specific component, follow the capitalization of the component name and apply monospace formatting, such as "the `Pod` spec", "a `Node` object", or "an `ImageStream` resource".
+
+*Use it*: yes
+
+*Incorrect forms*: operator, operatorize
+
+*See also*: link:https://github.com/openshift/openshift-docs/blob/main/contributing_to_docs/doc_guidelines.adoc#api-object-formatting[API object formatting] in OpenShift documentation guidelines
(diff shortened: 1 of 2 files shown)


bergerhoffer:
@redhat-documentation/ccs-style-council for review please!

supplementary_style_guide/glossary_terms_conventions/product_conventions/openshift.adoc
```diff
@@ -392,22 +392,43 @@ allowed for use across all distribution documentation:
 
 *See also*: xref:okd[OKD]
 
-[discrete]
 [[operator]]
 ==== Operator
 *Description*: An Operator is a method of packaging, deploying, and managing a
```
bergerhoffer, line 397: Instead of duplicating the information exactly, how about we just add a link to the general guidelines?

I feel like we might be setting ourselves up for failure if we duplicate it, and then update it one place but forget to do the other.

We do something like this already in a few places (even in the openshift file already).

```
[discrete]
==== node (noun)
See xref:node[node] in the _General Conventions_ part.
```

These duplicated ones for some reason don't have defined IDs (and they should), but we can take care of fixing that later.

supplementary_style_guide/glossary_terms_conventions/general_conventions/o.adoc
```diff
@@ -164,6 +164,51 @@
...
+While "containerized" is allowed, do not use "Operatorize" to refer to building
+an Operator that packages an application.
+
+NOTE: When referring generally to other Kubernetes components, such as pods, nodes, or image streams, use lowercase. When referring to a specific component, follow the capitalization of the component name and apply monospace formatting, such as "the `Pod` spec", "a `Node` object", or "an `ImageStream` resource".
+
+*Use it*: yes
+
+*Incorrect forms*: operator, operatorize
+
+*See also*: link:https://github.com/openshift/openshift-docs/blob/main/contributing_to_docs/doc_guidelines.adoc#api-object-formatting[API object formatting] in OpenShift documentation guidelines
```
bergerhoffer, line 210: Instead of pointing to our GitHub repo, I'd just use the entry we have on this already in this guide:

```
*See also*: xref:api-objects[API objects]
```

supplementary_style_guide/glossary_terms_conventions/general_conventions/o.adoc
```diff
@@ -164,6 +164,51 @@
...
+
+The term "Operator" in the context of Kubernetes is always capitalized to distinguish it from other types of operators, such as human or mathematical operators. For example:
+
+----
+= Support policy for unmanaged Operators
+
+Individual Operators have a `managementState` parameter in their configuration.
+----
+
+----
```
bergerhoffer, line 183: If you want to include an example of a place that uses a lowercase operator, I'd suggest adding a heading to each of these examples. Something like:

```
.Example: Kubernetes Operator
----
= Support policy for unmanaged Operators
Individual Operators have a `managementState` parameter in their configuration.
----

.Example: Mathematical operator
----
The following operators and operands are supported in Drools Rule Language:
* + (addition)
* - (subtraction)
...
----
```

**However** I'm wondering if it makes sense to break out the lowercase "operator" usage into its own term entry. Because it's slightly confusing to mix them here, because the "Incorrect forms" field lists "operator" as an incorrect entry.

bergerhoffer (commented):
Thanks for submitting this! Added some thoughts

supplementary_style_guide/glossary_terms_conventions/general_conventions/o.adoc
sterobin: Sure, I just thought the OpenShift doc guidelines provided a bit more helpful info, but fine by me. Done.

supplementary_style_guide/glossary_terms_conventions/product_conventions/openshift.adoc
sterobin: Yeah, I wasn't sure how we wanted to handle that, and didn't notice that example of cross-reffing that way. So yep, done. I've just cross-reffed in the same way.

supplementary_style_guide/glossary_terms_conventions/general_conventions/o.adoc
sterobin: Actually, on second thought, I recall now the other reason I referred to the OpenShift docs repo. This and the new xref I've just added to the OpenShift section per your suggestion won't work because they're xref's outside of the file. So it'll need to just be the repo link to the relevant section like I had, unless we want to get fancier with symlinking or a different way to xref outside of the file. I think the repo+section anchor link might be easiest. Whatever you prefer.

bergerhoffer: @sterobin Can you clarify how the xref won't work? If you look at the live version of the SSG, it's working right now.

For example: https://redhat-documentation.github.io/supplementary-style-guide/#_node_noun

That is in the fuse.adoc file, but xrefs to a term in the general conventions n.adoc file.

sterobin: @bergerhoffer Ah, okay, it's just not working locally then, and not in github. So if we're okay with this being functional only in the final ouptut but not in the repo preview etc., fine by me.

bergerhoffer: Right, it wouldn't work from GitHub. But you should be able to test it locally by doing `asciidoctor master.adoc` and viewing the generated `master.html` file.

inoxx03:
@redhat-documentation/ccs-style-council Please review this PR either provide an approval or suggest changes. Thank you!

inoxx03 (dismissed):

supplementary_style_guide/glossary_terms_conventions/general_conventions/o.adoc
```diff
@@ -164,6 +164,53 @@
...
+capitalized. If it includes a product name, defer the product's capitalization
+style guidelines. For example:
+
+- Red Hat OpenShift Logging Operator
+- Prometheus Operator
+- etcd Operator
+- Node Tuning Operator
+- Cluster Version Operator
+
+While "containerized" is allowed, do not use "Operatorize" to refer to building
```
ndeevy, line 203:
```suggestion
Although "containerized" is allowed, do not use "Operatorize" to refer to building
```
Suggesting this change because the ISG says don't use 'while' to mean 'though' or 'although'

supplementary_style_guide/glossary_terms_conventions/general_conventions/o.adoc
```diff
@@ -164,6 +164,53 @@
...
 
+[discrete]
+[[operator]]
+==== image:images/yes.png[yes] Operator (noun)
+*Description*: In the context of Kubernetes, an Operator is a method of packaging, deploying, and managing a
+Kubernetes application. A Kubernetes application is an application that is both
+deployed on a Kubernetes cluster (including OpenShift clusters) and managed
+using the Kubernetes APIs and `kubectl` or `oc` tooling.
+
+The term "Operator" in the context of Kubernetes is always capitalized to distinguish it from other types of operators, such as human or mathematical operators. For example:
```
ndeevy, line 175: Maybe because you're using the **.Example** label, you don' need the "For example:" here?

supplementary_style_guide/glossary_terms_conventions/general_conventions/o.adoc
```diff
@@ -164,6 +164,53 @@
 
 *See also*:
 
+[discrete]
+[[operator]]
+==== image:images/yes.png[yes] Operator (noun)
+*Description*: In the context of Kubernetes, an Operator is a method of packaging, deploying, and managing a
+Kubernetes application. A Kubernetes application is an application that is both
+deployed on a Kubernetes cluster (including OpenShift clusters) and managed
+using the Kubernetes APIs and `kubectl` or `oc` tooling.
```
ndeevy, line 173:
```suggestion
by the Kubernetes APIs and `kubectl` or `oc` tooling.
```

Suggesting this to remove the 'using'. See the ISG entry for 'using': "To avoid ambiguity, replace this gerund with either "by using" or "that uses"."  In this case, I think 'by' on its own might fix it? If that's not accurate, then it should be 'by using'.

ndeevy:
Reviewed @sterobin  a couple of small style comments added. Looks good otherwise.

supplementary_style_guide/glossary_terms_conventions/general_conventions/o.adoc
IngridT1: I agree, @ndeevy

supplementary_style_guide/glossary_terms_conventions/general_conventions/o.adoc
```diff
@@ -164,6 +164,51 @@
...
+
+* + (addition)
+* - (subtraction)
+...
+----
+
+An Operator's full name must be a proper noun, with each word initially
+capitalized. If it includes a product name, defer the product's capitalization
+style guidelines. For example:
+
```
IngridT1, line 196: Add *to* after "defer".

supplementary_style_guide/glossary_terms_conventions/product_conventions/openshift.adoc
```diff
@@ -392,22 +392,43 @@ allowed for use across all distribution documentation:
...
 
-_Install the etcd Operator._
+* + (addition)
+* - (subtraction)
+...
+----
+
+An Operator's full name must be a proper noun, with each word initially
+capitalized. If it includes a product name, defer the product's capitalization
+style guidelines. For example:
```
IngridT1: Add _to_ after "defer".

IngridT1 (changes requested):

supplementary_style_guide/glossary_terms_conventions/general_conventions/o.adoc
sterobin: Sure, although the syntactic ambiguity that IBM is trying to avoid only occurs when the object is followed by "using", which isn't the case here because it's a passive construction (and necessarily so). IBM even specifies this:

> To refer to something "using" one thing or another is often ambiguous. For example, the sentence "Back up the server using ABC" might mean "Back up the server by using ABC" or "Back up the server that uses ABC"

So in this example, "is managed using" causes no ambiguity because the object is at the start. Using "by using" here just makes it sound odd and clunky without the object, imo. Compared to "manage an application using" which would indeed need to be "by using".

supplementary_style_guide/glossary_terms_conventions/general_conventions/o.adoc
sterobin: Thanks, done.

supplementary_style_guide/glossary_terms_conventions/general_conventions/o.adoc
sterobin: Ah, good catch, done. Thanks.

supplementary_style_guide/glossary_terms_conventions/product_conventions/openshift.adoc
sterobin: See above.

inoxx03 (approved):

IngridT1 (dismissed):

supplementary_style_guide/glossary_terms_conventions/general_conventions/o.adoc
ndeevy: In this case, if there's no ambiguity then it's fair and logical to keep it as is. +1

ndeevy (approved):

sterobin:
@inoxx03 I rebased and pushed to resolve conflicts.

ndeevy:
LGTM!

Issue #124, opened 2021-10-06
Fix minor style issues and typos throughout the guide


inoxx03:
This issue tracks minor style issues and typos identified in the guide that can be fixed separately from larger updates.

* https://redhat-documentation.github.io/supplementary-style-guide/index.html#appliance-console
	* set-up -> setup (set-up not recommended by IBM SG, we use _setup_ throughouth the SSG)

* https://redhat-documentation.github.io/supplementary-style-guide/index.html#composite-content-view
	* The abbreviation for Composite Content View should probably be "CCV", not "CVV"

ndeevy:
#132 was merged for this issue. Closing.

Issue #83, opened 2021-05-19
"As expected"


mjahoda:
Because the IBM Style Guide (IMHO incorrectly) uses "as expected" a lot and I haven't found any written guideline regarding this, I am proposing to add something along the following to the Supplementary style guide:

"as expected" - expectations are relative and do not belong to technical documentation. Use "correctly" or "properly" instead.

jafiala:
I second this (although, full disclosure, I learned this from mjahoda).

I think the main argument against using "as expected" is that we cannot assume users' expectations. For example, a tester might _expect_ a task to fail, but that does not mean the task functions correctly.

jherrman:
This is definitely a fair point, though if I remember correctly, "as expected" is meant to be a more user-focused version of "as intended" (meaning to say the software does what it's bloody supposed to in a given situation). 

On top of that, you could find not-dissimilar issues with the meanings of "correctly" [1] and "properly" [2], too.

[1] I think what @jafiala points out is actually an argument _for_ using "as expected", since in scenarios where we talk about things failing/crashing (instead of e.g. risking hardware damage), "correctly" would be a strange term to use. In addition, there is the potential theme/rheme ambiguity of e.g. "program X does Y correctly" and "program X correctly does Y".

[2] "Proper" is a term strongly associated with a societal expectations, so "do something properly" can mean different things to different people.

So, overall, I'd be against this proposal, given that:

- "As expected" covers situations when "correctly" or "properly" would be stylistically awkward 
- "As expected" is already heavily used in our existing docs and removing it would mean overhead
- None of the terms at our disposal is 100% perfect anyway

mjahoda:
@jherrman As I am reading through your comment, I am regretting that I forgot to add "as intended" to this issue :-)

To your point [1] - I am not a native speaker and I do not see anything strange using "correctly" in the mentioned scenarios.

Regarding [2] - well, I also feel that "correct" is more appropriate than "properly" in technical docs

Speaking of the removal from all existing docs - I believe that having this as a recommendation and agreeing that there is no requirement for an immediate bulk action can resolve this concern.

ndeevy:
Hi!  my 2 cents:

I also dislike and remove "as intended/expected" but also for an additional reason: it's passive. Who or what expects/intends the behaviour? nobody knows usually.

[...]

PR #132, opened 2021-10-29
Fixing and updating minor style issues

ndeevy:
Hi all, this PR addresses issue: #124

It also applies IBM Style Guide and other conventions to various entries. Changes in this PR:

**IBM Style items**
- Changed 'set-up' to 'setup' in one case but replaced it with 'configuration' in one instance
- Removing future tense
- Removing possessive 's where it was applied to product or brand names
- Removing 'simply just'
- Replacing 'utilize'
- Changing 'once' to 'after'
- Addressing usages of the word 'using' to help translation and non-native readership
- Changing 'prior to' to 'before'
- Addressing a few passives for a more user-oriented construction (although there are mountains more! :-D)
- Replacing instances of a slash (/) between two items
- Replacing 'via' with 'through'
- Replacing 'may' with 'might'
- Replacing 'while' with 'although'
- Replacing the phrasal verb 'roll out'
- Rewording some anthropomorphic constructions
- Replacing 'desire'
- Replacing a few unnecessary parentheses with parenthetical commas

**Other conventions**

- Changing 'properly' for 'correctly' because this came up in a previous issue where 'proper' was less favorable than 'correct'. For discussion about that, see issue #83.

- Using the full name for OpenShift Container Platform. This guide itself has an entry that says to always use it when possible but didn't seem to use it consistently in the other entries. I'm not 100% sure about every instance though and can undo where I'm wrong.


supplementary_style_guide/glossary_terms_conventions/general_conventions/s.adoc
```diff
@@ -529,7 +529,7 @@
 [discrete]
 [[sudo]]
 ==== image:images/caution.png[with caution] sudo (noun)
-*Description*: `sudo` is a command that allows a user to run a program as another user (the root user by default). When a user requires elevated privileges, using the phrase 'as the root user' prior to a command is preferred over prefixing commands with `sudo`.
+*Description*: `sudo` is a command that allows a user to run a program as another user (the root user by default). When a user requires elevated privileges, use the phrase 'as the root user' before a command instead of prefixing commands with `sudo`.
```
bergerhoffer, line 532: CC @bburt-rh since you had volunteered to put together a PR for guidelines on commands that need elevated privileges.

Looks like we have some language here about using "as the root user" instead of prefixing commands with `sudo`. This is kind of the opposite guidance of what we were leaning toward in the meeting. We were leaning towards recommending to use `sudo` in front of the command, when possible, and if not, then make sure to say "as a root user" (not "the").

Not sure if we need to revisit based on this (I wasn't aware of this entry til now), or if we should move forward with as we discussed, and adjust this entry accordingly. Let me know what you think!

supplementary_style_guide/glossary_terms_conventions/product_conventions/openshift.adoc
```diff
@@ -215,7 +210,7 @@ The API object for a deployment can be either a Kubernetes-native `Deployment` o
 [discrete]
 [[image-stream]]
 ==== image:images/yes.png[yes] image stream (noun)
-*Description*: A series of Docker images identified by one or more tags. Image streams are capable of aggregating images from a variety of sources into a single view, including images stored in OpenShift’s integrated Docker repository, images from external Docker registries, and other image streams. The API object for an image stream is `ImageStream`.
+*Description*: A series of Docker images identified by one or more tags. Image streams are capable of aggregating images from a variety of sources into a single view, including images stored in the integrated Docker repository of the OpenShift Container Platform, images from external Docker registries, and other image streams. The API object for an image stream is `ImageStream`.
```
bergerhoffer, line 213: Can you remove "the" from in front of "OpenShift Container Platform"

bergerhoffer (commented):
Just one minor nit, but this looks great to me. Thanks a lot for taking a sweep through and correcting these issues!

supplementary_style_guide/glossary_terms_conventions/product_conventions/red-hat-cloudforms.adoc
```diff
@@ -3,7 +3,7 @@ For documentation questions, contact cloudforms-docs@redhat.com.
 [discrete]
 [[appliance-console]]
 ==== image:images/yes.png[yes] Appliance console (noun)
-*Description*: The appliance console is a command-line based interface into the Red Hat CloudForms appliance used for set-up and configuration.
+*Description*: The appliance console is a command-line based interface built into the Red Hat CloudForms appliance used for setup and configuration.
```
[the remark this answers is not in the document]
bergerhoffer: I'm not sure either. There is a cloudforms-docs@redhat.com email above that we could always consider running it past

supplementary_style_guide/glossary_terms_conventions/product_conventions/openshift.adoc
ndeevy:
```suggestion
*Description*: A series of Docker images identified by one or more tags. Image streams are capable of aggregating images from a variety of sources into a single view, including images stored in the integrated Docker repository of OpenShift Container Platform, images from external Docker registries, and other image streams. The API object for an image stream is `ImageStream`.
```

supplementary_style_guide/glossary_terms_conventions/general_conventions/s.adoc
bburt-rh: @bergerhoffer I didn't realize that this info was already here either. Here are couple of things I think we should consider: 
- Because there's only a single root user, "the root user" makes more sense to me than "a root user." 
- I think it's consistent with our discussion last week to recommend to "use the phrase 'as the root user' before a command" over using `sudo`. I like the phrasing here that states that it's a preferred practice rather than a required one because that leaves open the possibility of using `sudo` in some instances.
- That being said, I do think we might want to consider adding an extra sentence along the lines of  "Although you can also use a shell prompt as part of the command to indicate the level of privileges required to run the command  (`#` for root or `$` for regular user), never rely only on the shell prompt to inform users of the required privileges. Above all, use a consistent approach when giving instructions about issuing commands that require elevated privileges."

bergerhoffer: @bburt-rh I thought that the majority of people thought we ought to recommend using sudo when possible as the preferred option. And only if sudo didn't work, then to do the combined prompt + lead-in mention of being the root user.

And I think you're right about "the root user". I think I was misremembering. I believe the discussion was to not say "as root" but instead to say "as the root user".

@redhat-documentation/ccs-style-council please take a look at this, especially those that were in October's meeting where this issue was discussed. Please help us remember what the decision was, and what you think given this information that is already here in the sudo entry.

bergerhoffer (dismissed):
Thanks for the update @ndeevy. This LGTM!

supplementary_style_guide/glossary_terms_conventions/general_conventions/s.adoc
bburt-rh: @bergerhoffer  You're right. My own notes say that the guidance should be to use either `sudo` or an introductory phrase and not just to rely on a shell prompt. The consensus was that we wanted something more explicit than a hash mark or dollar sign. Maybe we should change the wording to reflect that we need something more than the shell prompt and that `sudo` or an introductory phrase are acceptable, depending on circumstance and what you need to do to stay consistent with the rest of the doc you're working on.

ndeevy:
> Just one minor nit, but this looks great to me. Thanks a lot for taking a sweep through and correcting these issues!

Thanks @bergerhoffer No problem. I'll wait for a few days to see if more folks have input on the 'sudo' entry before updating again.

bergerhoffer:
@ndeevy I think we can move forward with these updates and continue the sudo discussion / follow up separately if there are any updates needed for that.

@redhat-documentation/ccs-style-council Need a few more reviews/approvals please!

supplementary_style_guide/glossary_terms_conventions/product_conventions/ceph.adoc
```diff
@@ -512,7 +512,7 @@ See xref:container[container] in the _General conventions_ section.
 [discrete]
 [[snap]]
 ==== image:images/yes.png[yes] snap (noun)
-*Description*: A snap is the snapshot ID of an object. The only writable version of the object is called `head`. If an object is a clone, this field includes its sequential ID. Always mark it properly (`snap`).
+*Description*: A snap is the snapshot ID of an object. The only writable version of the object is called `head`. If an object is a clone, this field includes its sequential ID. Always mark it correctly (`snap`).
```
porueesq, line 515:
```suggestion
*Description*: A snap is the snapshot identifier of an object. The only writable version of the object is called `head`. If an object is a clone, this field includes its sequential identifier. Always mark it correctly (`snap`).
```

porueesq (changes requested):
Wow! Amazing job here. 
Adding my 2c for refinement.

supplementary_style_guide/glossary_terms_conventions/general_conventions/e.adoc
```diff
@@ -3,7 +3,7 @@
 ==== image:images/yes.png[yes] Emacs (noun)
 *Description*: "Emacs" is a text editor available in Unix. Like vi, Emacs is a screen editor. Unlike vi, Emacs is not an insertion mode editor, meaning that any character typed in Emacs is automatically inserted into the file, unless it includes a command prefix.
 
-If referring to the program, use "Emacs," for example, "Source-Navigator supports Emacs or vi commands." If referring to the shell prompt command, use `emacs`. The command should always be properly marked up, for example, "At the prompt, type `emacs`." The complete and correct name is "GNU Emacs."
+If referring to the program, use "Emacs," for example, "Source-Navigator supports Emacs or vi commands." If referring to the shell prompt command, use `emacs`. Always mark up the command correctly, for example, "At the prompt, type `emacs`." The complete and correct name is "GNU Emacs."
```
sterobin, line 6: @ndeevy PR looks good. Bringing up one thing only because I see above that you changed "correct." to "correct". and another instance of double quotation with punctuation. Your changes there are correct, but for a separate PR, we should ensure consistency with double quotes with other punctuation, like "word." vs. "word".  per [IBM guidance](https://www.ibm.com/docs/en/ibm-style?topic=punctuation-quotation-marks#quotation-marks-with-other-punctuation). The ruling is basically always "word". "word", etc. except when the quoted portion is a complete sentence on its own with its own punctuation, or if the component has its own punctuation.

This sample here is a perfect example of both cases: Everything in this line is exactly right, with cases of internal and external periods used perfectly, except the last instance, which should be "GNU Emacs".

Nothing you've done here of course, just pointing out the need for some alignment throughout based on the corrections you made previously. Something to consider.

sterobin (approved):
@ndeevy Good with me. Just one trivial comment beyond the scope of this PR. Thanks.

supplementary_style_guide/glossary_terms_conventions/general_conventions/e.adoc
ndeevy: Definitely something to consider, TY @sterobin  I've found times where I was unsure about this, which means others might too so it'd be good to have guidance.

ndeevy:
> Wow! Amazing job here. Adding my 2c for refinement.

Thanks for your cents, @porueesq    o/*

supplementary_style_guide/glossary_terms_conventions/product_conventions/red-hat-cloudforms.adoc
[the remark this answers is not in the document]
ndeevy: Mailed the team, waiting to hear back. In the meantime I could remove this change or merge and address in another PR if appropriate.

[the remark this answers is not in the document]
ndeevy: Haven't heard back so I think this PR might be safe to merge and I can adjust with a new PR if and when there's more guidance from that team.

ndeevy:
Thanks you all who reviewed this PR, I made some minor additional commits from feedback so it needs some fresh approvals to merge if you have time, please and thanks again.

@bergerhoffer 
@sterobin 
@porueesq 
@inoxx03

porueesq (approved):

bergerhoffer (approved):

PR #211, opened 2022-08-26
Added "IP addresses and MAC addresses"

mmuehlfeldRH:
In documentation people often use IP addresses from ranges reserved for private networks, such as 192.168.0.0/24. However, there are IP ranges exclusively reserved for documentation purposes to avoid conflicts and confusion. We should use them. Same for MAC addresses.

This enhancement also lists the RFCs that define the ranges.

### DIFF
--- a/supplementary_style_guide/style_guidelines/general-formatting.adoc
+++ b/supplementary_style_guide/style_guidelines/general-formatting.adoc
@@ -338,3 +338,16 @@ Avoid all references to the costs and charges of Red Hat products. Although the 
 .Example problematic phrase
 
 * "at no initial cost" - Avoid this phrase in documentation because, although it implies there are further costs, it can also be construed to mean that the product is free when it is not.
+
+[[ip-addresses-and-mac-addresses]]
+== IP addresses and MAC addresses
+
+Use the IP and MAC address ranges that are reserved for documentation purposes to avoid the likelihood of conflicts and confusion:
+
+* Reserved IP address ranges: See the _IBM Style Guide_ for the list of reserved IP address ranges to use in documentation.
+
+* Reserved MAC addresses (defined in link:https://www.rfc-editor.org/rfc/rfc7042.html#section-2.1.2[RFC 7042]):
+
+** For unicast: 00:00:5E:00:53:00 - 00:00:5E:00:53:FF
+** For multicast: 01:00:5E:90:10:00 - 01:00:5E:90:10:FF
+


supplementary_style_guide/style_guidelines/general-formatting.adoc
```diff
@@ -329,3 +329,38 @@ See xref:release-notes[Release notes] for guidelines about deprecation notices.
...
 
 image:images/yes.png[yes] It is anticipated that an upcoming release will include a fix for this issue.
+
+[[ip-addresses-and-mac-addresses]]
+== IP addresses and MAC addresses
+
+There are reserved ranges of IP and MAC addresses for documentation purposes. Use them to avoid the likelihood of conflicts and confusion.
+
+[discrete]
+=== Reserved IP address ranges
```
bergerhoffer, line 339: I believe that these items are already covered in the IBM style guide, so we wouldn't want to duplicate that guidance here in the SSG.

However if the MAC address piece isn't covered in the ISG, then I think it could make sense to add it in to the SSG.

Can you remove the section on IP addresses and keep just the MAC address piece. Or maybe we could leave this "Reserved IP address ranges" heading and just say to see the ISG for details about the reserved IP addresses to use in docs. Not sure which would be best - others on the council might have an opinion here.

mmuehlfeldRH: I replaced the part about IP addresses with a reference to the IBM Style Guide.

supplementary_style_guide/style_guidelines/general-formatting.adoc
```diff
@@ -329,3 +329,16 @@ See xref:release-notes[Release notes] for guidelines about deprecation notices.
 image:images/no.png[no] We will fix this issue in the 18.3.4 release next February.
 
 image:images/yes.png[yes] It is anticipated that an upcoming release will include a fix for this issue.
+
+[[ip-addresses-and-mac-addresses]]
+== IP addresses and MAC addresses
+
+Use the IP and MAC address ranges that are reserved for documentation purposes to avoid the likelihood of conflicts and confusion:
+
+* Reserved IP address ranges: See the `Using addresses in examples` section in the IBM Style Guide.
```
bergerhoffer, line 338: I don't think the other places we reference the ISG in here point to actual section titles. 

Can you change this to something like "See the _IBM Style Guide_ for the list of reserved IP address ranges to use in documentation."

bergerhoffer (commented):
Just one other thing, then this lgtm!

supplementary_style_guide/style_guidelines/general-formatting.adoc
bburt-rh: I agree with @bergerhoffer here. "Hard-coding" the section name in another guide isn't a good idea because the section name in the ISG could change, and we wouldn't want to have to track those possible changes. 

After this change, this LGTM as well.

dfitzmau: HI @mmuehlfeldRH . You OK to implement this change? I could give the third approval if it's done by today. I'll be going on PTO On Monday.

mmuehlfeldRH: I updated the reference.

bburt-rh:
lgtm

supplementary_style_guide/style_guidelines/general-formatting.adoc
bergerhoffer: Thanks!

bergerhoffer (approved):
LGTM!

bburt-rh (approved):

ncbaratta (approved):
lgtm

PR #232, opened 2022-09-28
Issue 220 - add style guidance for homographs

bburt-rh:
This PR adds a style entry with guidance for using homographs.

### DIFF
--- a/supplementary_style_guide/style_guidelines/grammar.adoc
+++ b/supplementary_style_guide/style_guidelines/grammar.adoc
@@ -83,5 +83,19 @@ As needed, adjust the conversational to _fairly conversational_ for an audience 
 
 Red Hat Enterprise Linux 8 delivers a stable, secure, and consistent foundation across hybrid cloud deployments with the tools needed to deliver workloads faster with less effort.
 
+[[homographs]]
+== Homographs
+A homograph is a word that is spelled the same as another word but has a different meaning when used as a different part of speech.
+Using homographs close together in a sentence or paragraph might confuse readers. 
+Therefore, be aware of this potential issue, and, when possible, avoid writing sentences that use homographs close to one another provided that you can do so without changing the technical meaning.
+
+The following list includes homographs that might commonly appear in technical documentation:
+
+* attribute
+* block
+* coordinates
+* number
+* object
+* project
 
 // TODO: Add new style entries alphabetically in this file


supplementary_style_guide/style_guidelines/grammar.adoc
```diff
@@ -83,5 +83,20 @@ As needed, adjust the conversational to _fairly conversational_ for an audience
...
 
+== Homographs
+A homograph is a word that is spelled the same as another word but has a different meaning, 
+often when the word is used as a different part of speech.
+Using homographs close together in a sentence or paragraph might confuse readers. 
+Therefore, when possible, avoid using homographs close to one another. 
+
+Homographs that might commonly appear in technical documentation include the following:
+
+* attribute
```
Amrita42, line 94: How about adding the following >> **object** (objects are used a lot in OpenShift docs e.g _A custom resource is an object that extends the Kubernetes API or allows you to introduce your own API into a project or a cluster_)  .  **File** ( We use YAML files but file has two diff meanings >> Every file on the same disk must have a different name OR
We began to file out into the car park.)  Manifest (e.g. https://docs.openshift.com/container-platform/4.11/operators/operator_sdk/osdk-pkgman-to-bundle.html#osdk-migrating-pkgman_osdk-pkgman-to-bundle)

supplementary_style_guide/style_guidelines/grammar.adoc
```diff
@@ -83,5 +83,20 @@ As needed, adjust the conversational to _fairly conversational_ for an audience
 
 Red Hat Enterprise Linux 8 delivers a stable, secure, and consistent foundation across hybrid cloud deployments with the tools needed to deliver workloads faster with less effort.
 
+== Homographs
+A homograph is a word that is spelled the same as another word but has a different meaning,
```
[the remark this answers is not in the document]
bburt-rh: Good suggestion. I'll make the change.

supplementary_style_guide/style_guidelines/grammar.adoc
bburt-rh: I agree with adding **object** and will add it. But I would argue that sentences with **file** (verb) and **manifest** (verb) are unlikely to appear in technical docs. What do others think?

aireilly: crash?

supplementary_style_guide/style_guidelines/grammar.adoc
```diff
@@ -83,5 +83,20 @@ As needed, adjust the conversational to _fairly conversational_ for an audience
...
+Homographs that might commonly appear in technical documentation include the following:
+
+* attribute
+* block
+* close
+* coordinates
+* minute
+* number
+* object
+* project
```
Srivaralakshmi, line 100: I do not agree to add an entry or a generic admonition to avoid the word "project". PTAL at my https://github.com/openshift/openshift-docs/pull/49201#discussion_r966092250 PR.

supplementary_style_guide/style_guidelines/grammar.adoc
Srivaralakshmi: Does the verb/adjective second fit in this list?

Srivaralakshmi: Does the verb/adjective second fit in this list?

bburt-rh: > Does the verb/adjective second fit in this list?

It's a homograph for sure, but imo both uses are not common in tech docs. I'm not sure how often we find "second" used a verb in technical writing. I'm trying to keep the list limited to homographs in which both uses of the word are somewhat common in technical docs.  Can you cite some uses where you've seen "second" used as a verb in tech docs? It's common in meeting minutes, for sure, ("I second that motion," e.g.) but outside of that, I don't really know how it'd be used in a technical context.

supplementary_style_guide/style_guidelines/grammar.adoc
Srivaralakshmi: Sure, suggest adding this statement: Writers should therefore be aware of this potential issue and seek alternatives if applicable and if possible, as long as it does not hinder or impact or alter the technical accuracy or interpretation.

Feel free to refine this as appropriate.

supplementary_style_guide/style_guidelines/grammar.adoc
bergerhoffer: This isn't going to be a comprehensive list, so I don't think we need to look for more examples to add. Just listing a few examples should be sufficient. I was almost thinking that the number of examples we have is pushing it (~5 would be good IMO). So I'd limit them to just a handful that we think is most common/likely.

supplementary_style_guide/style_guidelines/grammar.adoc
```diff
@@ -83,5 +83,20 @@ As needed, adjust the conversational to _fairly conversational_ for an audience
 
 Red Hat Enterprise Linux 8 delivers a stable, secure, and consistent foundation across hybrid cloud deployments with the tools needed to deliver workloads faster with less effort.
 
+== Homographs
```
bergerhoffer, line 86: Can you add a [[homographs]] ID?

supplementary_style_guide/style_guidelines/grammar.adoc
```diff
@@ -83,5 +83,20 @@ As needed, adjust the conversational to _fairly conversational_ for an audience
...
+A homograph is a word that is spelled the same as another word but has a different meaning when used as a different part of speech.
+Using homographs close together in a sentence or paragraph might confuse readers. 
+Therefore, when possible, avoid writing sentences that use homographs close to one another. 
+
+Homographs that might commonly appear in technical documentation include the following:
```
bergerhoffer, line 91: Per ISG should have something after "following". Putting a suggested rewrite below, but feel free to adjust however you see fit.

```suggestion
The following list includes examples of homographs that might appear in technical documentation:
```

bergerhoffer (commented):
Looking great! Left a few suggestions, plus I agree with taking a look at @Srivaralakshmi's suggestion and seeing how you can update the guidance to be more specific. I know you say "when possible", but this could be expanded to say about using an alternative as long as it doesn't change the technical meaning, etc.

supplementary_style_guide/style_guidelines/grammar.adoc
bburt-rh: Made an edit to emphasize this guidance as more a question of writer awareness than a hard and fast rule.

supplementary_style_guide/style_guidelines/grammar.adoc
bburt-rh: I agree @bergerhoffer. I've reduced the list to 6 of what I think might be the most common instances.

supplementary_style_guide/style_guidelines/grammar.adoc
bburt-rh: done

supplementary_style_guide/style_guidelines/grammar.adoc
bburt-rh: fixed

bburt-rh:
@bergerhoffer PTAL at the updates.

Amrita42:
looks good. I do have one suggestion >>Do we want to add some examples to make it more clear (like we have done for **Conscious Language** or **Conversational Style**.)

bergerhoffer (approved):
LGTM!

I don't have a strong opinion on whether adding an example would be helpful, or overkill. Leaving it up to your judgment!

Amrita42:
> @Amrita42 Thanks for the suggestion, but I think it'll be fine without an example. Leaving it as-is
 That's fine , I suggested so its optional .

Amrita42 (approved):

ndeevy (approved):

bergerhoffer:
3 project member approvals - merging!

PR #28, opened 2020-10-03
Added link guidance content

ndeevy:
Hi all,
Here is the recommendation for internal and external links from the *Best Practices for Internal Links* doc linked in the topics spreadsheet. To be honest I'm not sure how relevant all of it is to our needs and I think it needs more work so will be happy to get some help with it.
There were comments from the RH Storage & Partner Certification spreadsheet about the usefulness of all the info in the *Best Practices for Internal Links* doc so maybe we can revisit those concerns?

Nevertheless, for now I've done the following:
- Included most of it. 
- Cleaned up the writing for style.
- Removed the examples of good and bad links because there were many and I felt it was superfluous and maybe confusing to show people lots of incorrect examples. It's not prescriptive.
- Added linking example formatting. I did this because the best practices doc doesn't actually give exact guidance of what we can use across CCS for consistency. I added the format we have been implementing across OpenStack because it's IBM style compliant and fits documentation link needs.

Thanx!

PR #240, opened 2022-10-10
Added link guidance content

bergerhoffer:
PR that rebased #28 and will continue addressing feedback from that PR in future updates here.

Preview: http://file.rdu.redhat.com/~ahoffer/2022/main-links.html#links

### DIFF
--- a/supplementary_style_guide/style_guidelines/links.adoc
+++ b/supplementary_style_guide/style_guidelines/links.adoc
@@ -2,11 +2,43 @@
 [[links]]
 = Links
 
-// [[internal-links]]
-// == Internal links
+[[cross-references]]
+== Cross-references
 
-// [[external-links]]
-// == External links
+Follow these guidelines when adding cross-references within your documentation:
+
+* Include cross-references only when necessary.
+* If the information is critical, consider including it instead of cross-referencing.
+
+.Example AsciiDoc: Cross-reference
+----
+For more information about <topic>, see xref:<link>[<link_text>].
+----
+
+[[external-links]]
+== External links
+
+Follow these guidelines when linking externally:
+
+* Always include link text. Bare URLs (URLs without explanatory information) are unhelpful because they do not provide adequate context about the link target.
+* Use hyperlinks unless the URL is an example URL or is otherwise inaccessible to users. See the _IBM Style Guide_ for information about using addresses in examples.
+* By default, links are followable and crawlable. Do not use the "nofollow" link option unless absolutely necessary.
+* Avoid links to external sites when possible. External links can change or be unreliable.
+* Avoid deep links (to a specific page or image from another company) to prevent an ongoing maintenance responsibility and to prevent bypassing a website's legal information.
+
+.Example AsciiDoc: External link
+----
+For more information about <topic>, see link:<link>[<link_text>].
+----
+
+[[link-text]]
+== Link text
+
+Follow these guidelines when specifying link text:
+
+* Contextually describe what the user will find at the target location so that they can decide if they want to leave their current location.
+* Use a concise sentence or sentence fragment as the link text.
+* Avoid irrelevant link text.
 
 [[rh-kb-links]]
 == Links to Red Hat Knowledgebase articles


bergerhoffer:
@ndeevy Okay here is my attempt at wrapping up the link guidance from #28. Can you PTAL when you get a chance? 

I made the updates as separate commits so that you can see the progression of what changed compared to your initial PR. The second commit addresses some of the peer review feedback from the original PR. The third commit addresses the fact that we alphabetize the items in each category now. 

Here are a few of the significant changes I made:
* Updated the "internal link" section to be specifically about "cross-references". I was struggling with the idea of internal links meaning both cross-references and regular links to RH domain. So I thought it was better to get specific and keep this one just about xrefs. I got inspiration for this from Google's style guide: https://developers.google.com/style/cross-references
* Got rid of the general best practices section, because it was too hard to have a generic section, especially when I couldn't get it to come first alphabetically. I instead distributed the guidance to either the cross-references section or the external links section, as appropriate.
* Removed the references to not using JavaScript/iFrame/Flash links, because I don't even think this is possible in AsciiDoc

Let me know what you think and if there's anything you think I should revert or change. Thanks!

Preview: http://file.rdu.redhat.com/~ahoffer/2022/main-links.html#links

supplementary_style_guide/style_guidelines/links.adoc
```diff
@@ -2,11 +2,44 @@
...
+----
+For more information about <topic>, see xref:<link>[<link_text>].
+----
+
+[[external-links]]
+== External links
+
+Follow these guidelines when linking externally:
+
+* Use hyperlinks and use them consistently.
```
ndeevy, line 23: @bergerhoffer you commented in the original PR about this and I agree with you, it's not very helpful or clear what it's trying to say. Maybe just ' Use hyperlinks' ?

ndeevy (dismissed):

supplementary_style_guide/style_guidelines/links.adoc
bergerhoffer: Thanks @ndeevy! I adjusted this to say to use hyperlinks unless the URL is an example URL or is otherwise inaccessible. 

I also removed the hyperlink guidance from the cross-references section (because those would always be hyperlinks), and instead replaced it with some guidance on including info right there instead of xreffing if it's critical.

PTAL at these updates and let me know what you think! Diff is here: https://github.com/redhat-documentation/supplementary-style-guide/pull/240/commits/b6ed97fb477f1e4f089e471eb928f238c89486e4

ndeevy: Cleaner, clearer, awesome. LGTM

ndeevy (approved):

bergerhoffer:
@redhat-documentation/ccs-style-council This is a PR to wrap up the link guidance (originally in #28). PTAL when you can!

supplementary_style_guide/style_guidelines/links.adoc
```diff
@@ -2,11 +2,44 @@
...
+----
+For more information about <topic>, see xref:<link>[<link_text>].
+----
+
+[[external-links]]
+== External links
+
+Follow these guidelines when linking externally:
+
+* Avoid links to external sites when possible. External links can change or be unreliable.
```
bburt-rh, line 23: Suggestion: Maybe consider adding `When linking to external sites, use URLs with top-level domain names, if possible.` E.g. `https://prometheus.io` vs `https://prometheus.io/docs/prometheus/latest/management_api/`. I'm not sure this is necessary, but it might help reduce broken external links.

supplementary_style_guide/style_guidelines/links.adoc
```diff
@@ -2,11 +2,44 @@
...
+* Use hyperlinks unless the URL is an example URL or is otherwise inaccessible to users.
```
bburt-rh, line 24: Consider cross-referencing the ISG guidance on sample URLs: the "Using addresses in examples" section here: https://www.ibm.com/docs/en/ibm-style?topic=elements-web-ip-email-addresses

bburt-rh:
I added one suggestion to consider, but otherwise this lgtm. Thanks @bergerhoffer!

bburt-rh (approved):

supplementary_style_guide/style_guidelines/links.adoc
bergerhoffer: @bburt-rh So we don't actually ever link to the ISG from the SSG (since the SSG is public, and the ISG is not available to everyone outside of RH). We also avoid naming the sections too, but if you think it's worthwhile I could add something like 

```
See the _IBM Style Guide_ for information about using addresses in examples.
```

wdyt?

bergerhoffer: Thanks, updated!

bburt-rh (dismissed):

ndeevy (dismissed):

supplementary_style_guide/style_guidelines/links.adoc
```diff
@@ -2,11 +2,44 @@
...
+----
+
+[[link-text]]
+== Link text
+
+Follow these guidelines when specifying link text:
+
+* Contextually describe what the user will find at the target location so that they can decide if they want to leave their current location.
+* Use a concise sentence or sentence fragment as the link text.
+* Do not use bare URLs. Bare URLs are unhelpful because they do not provide adequate context about the link target.
```
bredamc, line 41: Perhaps we should define what is meant by "bare URL". I assumed that a bare URL is a link without link text, so I'm not sure if this point makes sense in this section, given that it's already included in the "External links" section? Perhaps it should be included in the Cross-references section instead?

bergerhoffer: @bredamc I actually thought about changing "bare URL" to "raw URL" because that's the term I'm more familiar with. wdyt? Or I could give an example if that would help.

I do think that the raw URL piece is more applicable in the external links section, since I don't even think it's possible to give a raw URL for cross-references.

bredamc: @bergerhoffer Apologies for not being more clear -- we should absolutely retain that point in the "External links" section. I meant that we should include it in the "Cross-references" section instead of in the "Link text" section, but I agree that there's no need if it's not possible to provide a raw URL in a cross-reference anyway! 

Whether we call it a bare URL or a raw URL, I think we should define what it is. When we say "Do not use bare URLs", do we mean "Always include link text"? If so, should we phrase it in that way for clarity? If we retain this point in the "Link text" section, should that text be moved to be the first point?

bredamc: @bergerhoffer Apologies for the delay.

I think we should rephrase the current "Bare URLs" text as follows:
"Always include link text. _Bare URLs_ (URLs without explanatory information) are unhelpful because they do not provide adequate context about the link target."

bredamc: Examples are always good -- similar to the picture that paints a thousand words ;)

bergerhoffer: I think your suggested edit makes it crystal clear, thanks! Updated

bredamc: One last tiny comment -- maybe reorder the list of points to list the "Do" statements first, and then the "Don't" statements?

bergerhoffer: Done!

bredamc (approved):
LGTM @bergerhoffer ! Many thanks for your patience :)

supplementary_style_guide/style_guidelines/links.adoc
```diff
@@ -2,11 +2,44 @@
...
+----
+
+[[external-links]]
+== External links
+
+Follow these guidelines when linking externally:
+
+* Avoid links to external sites when possible. External links can change or be unreliable.
+* Use hyperlinks unless the URL is an example URL or is otherwise inaccessible to users. See the _IBM Style Guide_ for information about using addresses in examples.
+* Default to followable, crawlable links. Do not use "nofollow" links unless absolutely necessary.
```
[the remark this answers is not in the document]
bburt-rh: These are relatively common HTML hyperlink concepts that have been around for awhile and are easy to find more info about (as you discovered!). I guess it depends on the level of HTML knowledge we're assuming about the writers using the SSG. Is it typical that the average tech writer working in a docs-as-code, Asciidoc workflow would know these concepts? My view is "yes." Maybe other people have a different view...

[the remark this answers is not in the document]
bredamc: I'm not familiar with those terms either, and would appreciate an explanation in the guide.

ndeevy (approved):

supplementary_style_guide/style_guidelines/links.adoc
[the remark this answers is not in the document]
ndeevy: This line is from the original guidance, which was for a much broader set of situations than docs. But I think you've got a point, Sharon and Breda, they're probably unknown to tech writers.
So maybe this line is not applicable to the documentation case really because we're being prescriptive about what to do in the rest of this guidance and writers are probably not in a position to use such types of links anyway? If that's the case, we could remove it.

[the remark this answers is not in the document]
bergerhoffer: So I just checked and apparently it is possible to define a link in AsciiDoc with a nofollow option. Would look something like this:

```
link:www.example.com[My example link, opts=nofollow]
```
So I think that this guidance is valuable to keep. I personally didn't know much about the difference of the two, other than knowing that you have to do something special to make a link "nofollow". 

However, I don't think we need to go too deep into explanations here. Maybe we could make it clearer that the default is followable/crawlable, and you'd have to do something special to enable "nofollow", and to not do that :)

What about something like this:

```suggestion
* By default, links are followable and crawlable. Do not use the "nofollow" link option unless absolutely necessary.
```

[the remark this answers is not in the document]
bredamc: Your suggestion LGTM @bergerhoffer!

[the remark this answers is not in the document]
bergerhoffer: Thanks @bredamc, I pushed the change (so that if @bburt-rh and @sbmetz agree, this can be approved if otherwise ready).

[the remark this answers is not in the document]
ndeevy: This makes sense!

bredamc (approved):
LGTM :)

supplementary_style_guide/style_guidelines/links.adoc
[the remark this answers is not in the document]
sbmetz: That works for me! Thanks!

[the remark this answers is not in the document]
bergerhoffer: Thanks all! @sbmetz would you mind approving the PR officially if it looks okay to you? (You go to files changed tab -> review changes -> hit the 'approve' radio button)

bburt-rh:
Looks great! Thanks @bergerhoffer for all the iterations. LGTM.

bburt-rh (approved):

bergerhoffer:
Has the necessary approvals, merging!

Issue #213, opened 2022-08-30
Can we revisit the word-usage guidance on plugin (noun)? 


michelle-purcell:
![image](https://user-images.githubusercontent.com/92924207/187479099-71ee86d8-b87e-4970-99d6-9d9c3cb3fe58.png)

**Problem:** 

“Plugin” and “plug-in” (noun) are used synonymously in published technical documentation online.

Red Hat and IBM styles favor the hyphenated form, whereas many of our other integrated technologies and developer-type community projects prefer 'plugin' (see google example below). This creates a lot of churn and pushback in our upstream communities. Many developers prefer 'plugin' without the hyphen. 

Our current style guide advice to use 'plug-in' instead of 'plugin' was rejected by our Quarkus upstream project community engineers. Our developers prefer 'plugin'. This creates challenges and lots of churn for Red Hat writers who have to convert 'plugin' (used widely upstream) to 'plug-in' (required downstream), especially as we move to an upstream-first content development model. It's doable but our time could be better spent on other content quality enhancements.

While 'plug-in' is said to be more grammatically correct, according to the Merriam-Webster dictionary, it also seems to be seen as a more formal and 'old-fashioned' form of the word. Is it time to relax the rule, especially for developer-type content, where the plugin tends to trump plug-in? Could we revisit the word usage for 'plug-in' with the IBM Style council?

**Example:**
https://developers.google.com/style/word-list#letter-p

![image](https://user-images.githubusercontent.com/92924207/187473279-7f24b16d-0499-4eed-82a1-7ec9ce2e84da.png)

lahinson:
@michelle-purcell The co-chairs of the IBM Style Council meet periodically with a few writers and editors from Red Hat to discuss issues like this one. I think the next meeting is 9 September. Would you like me to add this item to the agenda? I can also ask for an invitation for you to join, if you like.

michelle-purcell:
@lahinson - Thank you, Laura. It would be great if you could raise this. I will try to attend the discussion if I can. Do let me know if you need any further information or examples. :-)

bredamc:
@michelle-purcell As a workaround, you could replace the problem word with a {plugin} attribute, and define that attribute to have the value "plugin" upstream and "plug-in" downstream?

michelle-purcell:
@bredamc - Thank you Breda for the great idea. 

[...]

PR #253, opened 2022-10-28
Added new entries for plugin and plug-in

michelle-purcell:
Issue #213.

### DIFF
--- a/supplementary_style_guide/glossary_terms_conventions/general_conventions/p.adoc
+++ b/supplementary_style_guide/glossary_terms_conventions/general_conventions/p.adoc
@@ -259,6 +259,41 @@ When combined with privileges, permissions define a user's or resource's overall
 
 *See also*:
 
+[discrete]
+[[plugin-adj]]
+==== image:images/yes.png[yes] plugin (adjective)
+*Description*: Use to distinguish software code separate from the core application or service that adds new features or extends the functionality. 
+Use "plugin" rather than "plug-in", unless you are updating existing content that uses the hyphenated form.
+
+*Use it*: yes
+
+*Incorrect forms*: plug in, plug-in
+
+*See also*: xref:plugin[plugin]
+
+[discrete]
+[[plugin]]
+==== image:images/yes.png[yes] plugin (noun)
+*Description*: A _plugin_ is a software component that adds new features or extends the functionality of an existing application or service.
+Use "plugin" rather than "plug-in", unless you are updating existing content that uses the hyphenated form.
+
+*Use it*: yes
+
+*Incorrect forms*: plug in, plug-in
+
+*See also*: xref:plug-in[plug-in]
+
+[discrete]
+[[plug-in]]
+==== image:images/caution.png[with caution] plug-in (noun)
+*Description*: Use with caution. Write as shown only when updating existing content that uses the hyphenated form. For new content, use "plugin".
+
+*Use it*: with caution
+
+*Incorrect forms*: plug in
+
+*See also*: xref:plugin[plugin]
+
 // OCP: Added "In Kubernetes," and removed first sentence
 [discrete]
 [[pod]]


supplementary_style_guide/glossary_terms_conventions/general_conventions/p.adoc
```diff
@@ -259,6 +259,29 @@ When combined with privileges, permissions define a user's or resource's overall
 
 *See also*:
 
+[discrete]
+[[plugin]]
+==== image:images/yes.png[yes] plugin (noun)
+*Description*: A "plugin" is a software component that adds new features or extends the functionality of an existing application or service. 
+In Red Hat, "plugin" is preferred over "plug-in".
```
bburt-rh, line 266:
```suggestion
In Red Hat product documentation, "plugin" is preferred over "plug-in". 
```
- Beginning of the the sentence needs something more than `In Red Hat`, imo.
- Also, is `plugin` just "preferred"? Or do we want to be more opinionated here? Maybe something like `Use "plugin" rather than "plug-in". To my ears, using`preferred` leaves it open to use either form, though "plug-in" is immediately listed as an "incorrect" form and we say: "Use it: no".

michelle-purcell: @bburt-rh Thank you for reviewing :-) Great points. I have rephrased the sentence in line 266 to:

`In Red Hat product documentation, use "plugin" rather than "plug-in".`

bburt-rh: @michelle-purcell Or consider dropping the introductory phrase entirely. We're already in the SSG so we already know we're referring to RH product docs.

michelle-purcell: @bburt-rh - +1 - The fluff is now removed!

sbmetz: Do we need a separate entry for plug-in just to refer to plugin? 
Also, do we need to cover plug(-)in (adjective)?

sbmetz:
Do we need a separate entry for plug-in just to refer to plugin?
Also, do we need to cover plug(-)in (adjective)?

michelle-purcell:
@sbmetz - Thank you for reviewing.

> Do we need a separate entry for plug-in just to refer to plugin? 
It might be helpful to have a separate entry because the hyphenated noun form was previously preferred and is currently widely used in Red Hat documentation. Secondly, we deviate from IBM on the word usage guideline for 'plug-in'. Happy to remove if y'all think this is overkill.

> Also, do we need to cover plug(-)in (adjective)?
Good point. I can add another entry for the adjective form of "plugin" too, for example:

```
[discrete]
[[plugin-adj]]
==== image:images/yes.png[yes] plugin (adjective)
*Description*: Use to distinguish software code separate from the core application or service that adds new features or extends the functionality. 
Use "plugin" rather than "plug-in".

*Use it*: yes

*Incorrect forms*: plug in, plug-in

*See also*: xref:plugin[plugin]
```

What do you think?

sbmetz:
Since plugin and plug-in would appear next to each other, I think it's probably overkill. Open to others' input/opinions, of course. Adjective entry looks good :-)

bergerhoffer:
> Since plugin and plug-in would appear next to each other, I think it's probably overkill. Open to others' input/opinions, of course. Adjective entry looks good :-)

I'd recommend that we do something similar to what the ISG does for when guidance changes. Here's an example of "user name" (old guidance) and "username" (new guidance):

image

So my recommendation would be to keep the "plug-in" entry, but instead of "no", say "with caution". And say to only use it when updating existing content that uses that form. And for new content, to use "plugin". wdyt?

bburt-rh:
I agree with the suggestion made by @bergerhoffer. It makes sense in this case because we are only making this change with new content.

michelle-purcell:
@sbmetz @bburt-rh @bergerhoffer - Thanks so much for your input. I submitted updates to resolve the suggestions and comments above. WDYT? 

Do we need to advise writers to ensure that each content collection uses either plugin or plug-in consistently across the entire doc set or is that overkill?

bburt-rh:
> Do we need to advise writers to ensure that each content collection uses either plugin or plug-in consistently across the entire doc set or is that overkill?

@michelle-purcell Because we aren't applying the change retroactively to existing docs, I think it's not necessary. If we told people to ensure that each content collection uses one or the other consistently, then I think we'd be implying that they need to go back and change all existing instances in any doc set they're touching. Does that make sense?

bergerhoffer:
> > Do we need to advise writers to ensure that each content collection uses either plugin or plug-in consistently across the entire doc set or is that overkill?
> 
> @michelle-purcell Because we aren't applying the change retroactively to existing docs, I think it's not necessary. If we told people to ensure that each content collection uses one or the other consistently, then I think we'd be implying that they need to go back and change all existing instances in any doc set they're touching. Does that make sense?

I agree. I think that saying hey, we prefer "plugin" now, but it's fine to keep using "plug-in" if you're updating existing content is enough. I think that leaves flexibility for each team to decide how they want to implement the change. Maybe they want to do it little by little, or maybe they want to do an entire sweep of the doc set.

supplementary_style_guide/glossary_terms_conventions/general_conventions/p.adoc
```diff
@@ -259,6 +259,41 @@ When combined with privileges, permissions define a user's or resource's overall
...
+*Incorrect forms*: plug in, plug-in
+
+*See also*: xref:plug-in[plug-in]
+
+[discrete]
+[[plug-in]]
+==== image:images/caution.png[with caution] plug-in (noun)
+*Description*: Use with caution. Write as shown only when updating existing content that uses the hyphenated form. For new content, use "plugin".
+
+*Use it*: with caution.
```
bergerhoffer, line 291: Can you remove this period?

bergerhoffer (commented):
One tiny thing, otherwise LGTM!

supplementary_style_guide/glossary_terms_conventions/general_conventions/p.adoc
```diff
@@ -259,6 +259,41 @@ When combined with privileges, permissions define a user's or resource's overall
...
+*Use it*: yes
+
+*Incorrect forms*: plug in, plug-in
+
+*See also*: xref:plugin[plugin]
+
+[discrete]
+[[plugin]]
+==== image:images/yes.png[yes] plugin (noun)
+*Description*: A "plugin" is a software component that adds new features or extends the functionality of an existing application or service.
```
bergerhoffer, line 277: Sorry noticed one last thing. "plugin" should be in italics here, since we're defining the term

michelle-purcell: @bergerhoffer - Thanks for flagging. Fixed!

supplementary_style_guide/glossary_terms_conventions/general_conventions/p.adoc
michelle-purcell: Done!

sbmetz (approved):

bburt-rh (approved):

bburt-rh:
Looks good to me. Thanks for all of your work on this, @michelle-purcell!

bergerhoffer (approved):
LGTM!

bergerhoffer:
3 approvals, merging.

Thanks again @michelle-purcell for submitting this PR, and for putting the issue forward in the first place!

PR #257, opened 2022-11-15
Removing accidental preview file

bergerhoffer:
Accidentally added a preview file in #251. Removing it now!

PR #251, opened 2022-10-26
Adding link to stylepedia style guide

bergerhoffer:
Per #236, linking to the stylepedia style guide, in case folks outside of CCS land on our SSG.

Preview: http://file.rdu.redhat.com/~ahoffer/2022/main-stylepedia.html#_about_this_guide

@daobrien to add a similar link in the stylepedia landing page and guides to link to the SSG for CCS folks.

PR #258, opened 2022-11-15
Updated .gitignore to prevent future accidental preview inclusions

bergerhoffer:
Also related to #251, updated .gitignore so that I don't accidentally include other preview files when I rename them to share the previews.

Had to do separately from #257 otherwise it wouldn't have let me remove that file.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,2 @@
 master.html
-main.html
+main*.html


bredamc (approved):
LGTM :)

bergerhoffer:
Thanks! Merging before waiting for 3 approvals since this is just maintenance misc.

Issue #206, opened 2022-08-18
Add guidance for the correct capitalization and short form for referring to basic authentication


michelle-purcell:
### Problem
In Red Hat customer-facing documentation the term "basic authentication" is capitalized inconsistently as outlined in the examples below. Guidance is needed for when to use or not use each variation.

Lowercase **_basic authentication_** is the most commonly used across Red Hat docs, especially when referring to the authentication method in general. All lowercase is also preferred by the CCS style council in this gChat discussion: https://chat.google.com/room/AAAAAjCmTJI/AVRNWBJ3bwI

However, SMEs for Quarkus have pushed back on the lowercase form and prefer **_Basic authentication_**, when specifically referring to authentication that is based on the ['Basic' HTTP Authentication Scheme' RFC specification](https://www.rfc-editor.org/rfc/rfc7617.txt), which uses **_Basic authentication_** throughout the definition. 
![image](https://user-images.githubusercontent.com/92924207/185329030-8f9e554c-2a4e-4080-a9d6-93bc1f20b81d.png)

**BA** is aso a commonly used short form for the term.

### Request
Provide guidance in the style guide for the correct capitalization and also guidance for using the initialism.

### Examples of inconsistency:

**_basic authentication_**

![image](https://user-images.githubusercontent.com/92924207/185319191-e5ed5f54-8414-4f03-acaa-9078ca06d77c.png)
Source: https://access.redhat.com/documentation/en-us/red_hat_insights/2022/html/configuring_basic_authentication_for_red_hat_insights/con-insights-client-config-basic-auth-introduction_insights-client-config#con-insights-client-config-basic-auth-when-to_insights-client-config

**_Basic authentication_**

![image](https://user-images.githubusercontent.com/92924207/185317848-ef2e83ff-9894-4e4f-8be6-2053f548e3e3.png)
Source: https://access.redhat.com/documentation/en-us/cost_management_service/2022/html/adding_an_openshift_container_platform_source_to_cost_management/assembly-adding-openshift-container-platform-source#installation-overview_adding-an-ocp-source

**_Basic Authentication_**

![image](https://user-images.githubusercontent.com/92924207/185331512-ce5f448b-8979-423e-88bb-199d500b3ee0.png)
Source: https://www.ibm.com/docs/en/sva/7.0.0?topic=processing-configuring-basic-authentication

[...]

PR #271, opened 2023-01-12
206-guidance-for-basic-authentication

bburt-rh:
Summary: This PR addresses #206 by adding guidance for usage of `Basic authentication`.

### DIFF
--- a/supplementary_style_guide/glossary_terms_conventions/general_conventions/b.adoc
+++ b/supplementary_style_guide/glossary_terms_conventions/general_conventions/b.adoc
@@ -79,6 +79,19 @@
 *See also*: xref:distinguished-name[distinguished name]
 
 [discrete]
+[[basic-http-authentication]]
+==== image:images/yes.png[yes] Basic HTTP authentication (noun)
+*Description*: _Basic HTTP authentication_, or _Basic authentication_, is an authorization scheme that uses HTTP to send login credentials encoded in Base64. 
+This scheme is often used by applications to authenticate with servers, services, and API endpoints. 
+On first use, use "Basic HTTP authentication"; thereafter, use "Basic authentication".
+
+*Use it*: yes
+
+*Incorrect forms*: basic authentication, Basic Authentication, basic auth, Basic auth, Basic Auth
+
+*See also*:
+
+[discrete]
 [[basically]]
 ==== image:images/no.png[no] basically (adverb)
 *Description*: _Basically_ is another term for _in principle_ or _fundamentally_.


supplementary_style_guide/glossary_terms_conventions/general_conventions/b.adoc
```diff
@@ -78,6 +78,19 @@
 
 *See also*: xref:distinguished-name[distinguished name]
 
+[discrete]
+[[basic-http-authentication]]
+==== image:images/yes.png[yes] Basic HTTP authentication (noun)
+*Description*: _Basic HTTP authentication_, or _Basic authentication_, is an authorization scheme that uses Hypertext Transfer Protocol (HTTP) to send login credentials as user ID/password pairs that are encoded in Base64.
```
bergerhoffer, line 84: Not critical, but I don't think you need to spell out HTTP. It's listed as one of the easy known abbreviations (and we don't spell it out anywhere else in the SSG).

bergerhoffer (commented):
One minor comment, otherwise LGTM!

supplementary_style_guide/glossary_terms_conventions/general_conventions/b.adoc
mportman12: Agree with @bergerhoffer

mportman12 (commented):
Looks good except the small HTTP fix suggested by Andrea.

supplementary_style_guide/glossary_terms_conventions/general_conventions/b.adoc
bburt-rh: Thanks! Fixed.

mportman12 (approved):
LGTM

supplementary_style_guide/glossary_terms_conventions/general_conventions/b.adoc
```diff
@@ -78,6 +78,19 @@
 
 *See also*: xref:distinguished-name[distinguished name]
 
+[discrete]
+[[basic-http-authentication]]
+==== image:images/yes.png[yes] Basic HTTP authentication (noun)
```
bredamc, line 83: Should we add another entry for "Basic authentication", with each entry having a "See also" link to the other, similar to the "AMQ" and "Red Hat AMQ" entries? Or would that be overkill when the entries are sequential?

Maybe instead include both entries in the term, similar to the "asap, ASAP" entry in the _IBM Style_ word usage topic? 

In either of the above cases, we should list them alphabetically by letter (ignoring case): "Basic authentication, Basic HTTP authentication" -- this approach has the advantage of listing the most likely search term first?

bredamc: Just to clarify, the above comments are for the term only, not the term definition -- I agree with the current Description text which mentions the longer term first  :)

supplementary_style_guide/glossary_terms_conventions/general_conventions/b.adoc
```diff
@@ -78,6 +78,19 @@
...
+*Description*: _Basic HTTP authentication_, or _Basic authentication_, is an authorization scheme that uses HTTP to send login credentials as user ID/password pairs that are encoded in Base64.
```
bredamc, line 84: Should we try to rephrase this text to avoid the slash?

supplementary_style_guide/glossary_terms_conventions/general_conventions/b.adoc
bredamc: Just realised that if we have two entries, they won't actually be sequential -- the "basically" entry would separate them :)

bergerhoffer (approved):
LGTM, thanks!

supplementary_style_guide/glossary_terms_conventions/general_conventions/b.adoc
bburt-rh: I think it's fine to keep it simple and leave it as-is. But I'd like to hear what others think about the suggestion and am open to changing it if that's the consensus.

supplementary_style_guide/glossary_terms_conventions/general_conventions/b.adoc
bburt-rh: Agree. I'm just going to solve the problem by cutting that bit out as follows `to send login credentials that`...

supplementary_style_guide/glossary_terms_conventions/general_conventions/b.adoc
bergerhoffer: I don't have a strong opinion here. I think we do sometimes have both as entries, but not always. I'm okay with just the one entry.

bredamc (approved):
LGTM

bergerhoffer (approved):
LGTM

supplementary_style_guide/glossary_terms_conventions/general_conventions/b.adoc
bburt-rh: To keep things simple, I'm leaving this as a single entry.

mportman12 (approved):
LGTM

Issue #273, opened 2023-01-13
Terms addition: Kubernetes


jeana-redhat:
Didn't find guidance on proper use of "Kubernetes" in the OCP glossary, IBM word list, or here. Figured SSG is a pretty good home for it :)

Do use: Kubernetes
Do not use: kubernetes, k8s, `k8s`

bergerhoffer:
I'll work on an entry for this

PR #277, opened 2023-01-18
GitHub-273: Adding Kubernetes to glossary

bergerhoffer:
Addresses #273.

Preview: http://file.rdu.redhat.com/~ahoffer/style_guide/main-kubernetes.html#kubernetes


bergerhoffer:
@jeana-redhat can you review to see if this looks like what you were expecting?

jeana-redhat (approved):
Oh, with a definition too - even better than I was asking for :) Thanks!

mportman12 (approved):
LGTM

bergerhoffer:
> This looks good overall. Do you think it would also be good to mention "OKD" and "OpenShift" here along with cross-references to those entries?

@bburt-rh I briefly considered adding something about OpenShift, but I didn't want to add something that I wasn't 100% confident was accurate, so I avoided it. However, I think it's safe to add them as cross-references in the see also section, so I will do that.

bburt-rh (approved):

bburt-rh:
Thanks @bergerhoffer. Looks good to me. Approved!

ncbaratta (approved):
lgtm

PR #280, opened 2023-01-20
Small fix for Vale rules generation

aireilly:
Small fix to facilitate the Vale rules generation in https://github.com/redhat-documentation/vale-at-red-hat/pull/410

### DIFF
--- a/supplementary_style_guide/glossary_terms_conventions/general_conventions/o.adoc
+++ b/supplementary_style_guide/glossary_terms_conventions/general_conventions/o.adoc
@@ -414,7 +414,7 @@ NOTE: When referring generally to other Kubernetes components, such as pods, nod
 
 [discrete]
 [[operator-lifecycle-manager]]
-==== image:images/yes.png[yes] Operator Lifecycle Manager  (noun)
+==== image:images/yes.png[yes] Operator Lifecycle Manager (noun)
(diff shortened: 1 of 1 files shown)


joaedwar (approved):
LGTM

bredamc (approved):
LGTM :)

ncbaratta (approved):
lgtm

aireilly:
Thanks all!

supplementary_style_guide/glossary_terms_conventions/general_conventions/o.adoc
```diff
@@ -479,9 +479,9 @@ Use Organization Administrator as a proper noun when referring to the Organizati
 
 *Use it*: yes
 
-*Incorrect forms*: xref:ceph-osd[ceph-osd], xref:object-storage-device[Object Storage Device], xref:osd-daemon[OSD daemon]
+*Incorrect forms*: ceph-osd, Object Storage Device, OSD daemon
```
bergerhoffer, line 482: I know I'm late to the party here, but I'm thinking that these items were supposed to be in the "See also" line, and not in the incorrect forms section at all.

If you look at these other places this entry links to, that's how they all look. I'm thinking we should remove them here in a new PR. What do others think?

aireilly: Sure - but the "incorrect forms" line is still correct here right?

bergerhoffer: I think it needs to be empty. I think these xrefs were mistakenly put in the "Incorrect forms" field instead of the "See also" field.

aireilly: @ndeevy any idea about the above?

Issue #224, opened 2022-09-22
Man page references in "additional resources"


mmuehlfeldRH:
Do we have a CCS-wide rule how to write man page references in "additional resources" sections?

* `sudoers(5)`
* `sudoers(5)` man page
* The `sudoers(5)` man page
* ...

Minimalists write only the man page name and chapter number. Other writers append "man page" to make clear what it is. However, at least for me, the latter feels incomplete without article. ...and we have more styles in our docs.

It seems we don't have an official rule yet. Let's discuss it and add the result to the RH SSG.

PS: The same question applies also to other references in "additional resources", for example for files: (`/etc/fstab`, `/etc/fstab` file, The /etc/fstab` file,...)

bergerhoffer:
I don't have personal experience with referencing man pages in additional resources section, but I think that specifying "man page" is important. And that if you're going to have that, then the "the" in front makes sense. So my vote would be for the third option of:

* The `sudoers(5)` man page

bburt-rh:
+1 to what @bergerhoffer  says. 

I think we should add this guidance (3rd option) to the SSG. There's nothing about man pages in the ISG, btw.

mnsteven:
I also agree with @bergerhoffer, but that may well be because this is the style I used in previous companies.  The issue for me is using literals as nouns, so I also would do the same thing for files (the `/etc/lvm.conf` file).

This just came up in my release notes, where many of my headings refer to a feature of the `pcs` command-line interface.  So I had consistently used a title like "The pcs command-line interface now supports...".  All three of my peer reviewers said to use just "pcs" rather than "the pcs command-line interface".  It definitely makes for a shorter title, and I went with it, but I don't think it's correct -- although it doesn't bother me quite so much in a title as in a body of text.

bergerhoffer:
So one thing I just thought of is that we had a similar issue opened a few months ago: #169. The IBM style guide (ISG) does provide some generic information on things like this (about specifying what the items is, "man page" in our case) after it. 

We didn't end up adding any guidance to the SSG for this raised issue, because the ISG provided ample guidance, and we don't duplicate guidance. I'm wondering if that can extend enough to cover your man page question here.

The examples given in the sections, do cover also including the "the". Though technically these are full sentences, which yours wasn't. From the ISG:

> Generally, when you specify the name of a command, parameter, or option in text, specify the word command, parameter, or option after the name.
> ...
> (incorrect) You can use db2trc to record information about operations.
[...]

PR #288, opened 2023-02-08
Adding guidance for referencing man pages

bergerhoffer:
Draft for addressing #224.

Preview: https://file.rdu.redhat.com/~ahoffer/2024/main-man-pages.html#man-pages

### DIFF
--- a/supplementary_style_guide/style_guidelines/formatting.adoc
+++ b/supplementary_style_guide/style_guidelines/formatting.adoc
@@ -8,6 +8,23 @@ Follow the _IBM Style_ guide advice of _day Month year_ for date formats, for ex
 
 When the format _day Month year_ causes a presentation or clarity issue, use _Month day, year_ (for example, October 3, 2019) instead.
 
+[[man-pages]]
+== Man page references
+
+When referencing a man page in an "Additional resources" section, use the following format:
+
+* `_<man_page_name>_(_<section_number>_)` man page on your system
+
+Do not link to a website that contains the man page information. The contents of a man page might vary between systems or package versions, so users must run the `man` command on the target system to view the system-specific information for the named command or utility.
+
+.Example: Single man page
+
+* `sudoers(5)` man page on your system
+
+.Example: Multiple man pages
+
+* `nmcli(1)`, `nm-settings(5)`, and `sudoers(5)` man pages on your system
+
 [[non-breaking-spaces]]
 == Non-breaking spaces
 


bergerhoffer:
@redhat-documentation/ccs-style-council Draft is ready for adding guidance for how to reference man pages. 

My only outstanding questions are:

1. Is it okay to add something to the additional resources sections that is not a link? Or is a link a true requirement? @emmurphy1 is this something you can weigh in on?
2. Thoughts on if "Formatting" is the right section that this should go under?

supplementary_style_guide/style_guidelines/formatting.adoc
```diff
@@ -8,10 +8,21 @@ Follow the _IBM Style_ guide advice of _day Month year_ for date formats, for ex
 
 When the format _day Month year_ causes a presentation or clarity issue, use _Month day, year_ (for example, October 3, 2019) instead.
 
+[[man-pages]]
+== Man page references
+
+When referencing a man page, use the following format: `<name_of_utility>(<section_number>)` man page.
```
mportman12, line 14:
```suggestion
When referencing a man page, use the following format: _`<name_of_utility>(<section_number>)` man page_.
```
Italicize the template text to make it clear what is part of the template and what isn't. This formatting is used for date formats.

mportman12 (changes requested):
Formatting section is a good location for this. Left a formatting recommendation, otherwise looks good.

supplementary_style_guide/style_guidelines/formatting.adoc
bergerhoffer: Thanks, updated!

bergerhoffer: Actually one sec, that messed up the backtick formatting.

bergerhoffer: Okay, updated

bburt-rh (dismissed):

mportman12 (dismissed):
LGTM

ncbaratta (approved):
lgtm

emmurphy1:
@bergerhoffer   I checked the templates and ref guide and we don't say that you can reference man pages in additional resources, but I think they should be allowed. Let me get a discussion going on the the mod docs channel and I'll update here.

bergerhoffer:
Putting this PR on hold until we hear the result from the mod docs team!

soulcharmer:
I have one more issue to discuss.

When referencing a man page in the docs on the CP, the page is not readily available to the user. It appears that often, the user concludes that they need to google the man page, especially if they just browse the docs and do not have a subscription yet. We even got several bug reports to that effect, most recently this one: https://bugzilla.redhat.com/show_bug.cgi?id=2158027 This is a gap and the UX is also not that great because it is reasonable to expect to have the man page readily available on the CP.

However, in some products, there are big technical and logistical challenges to generate a stand-alone man page portal that would close this gap.

At the same time, the man page references are actually meant to point to the ones the user has installed on their system. If this detail is not explicit in the docs though, it can lead to bug reports or affect the UX...

A fairly simple workaround could be to set up a guideline to refer to man pages in the docs in a certain way and with the above in mind.

Thus far, we brainstormed two possible solutions in our team:

1) mention the installed system explicitly in the reference snippet, for example "For details, see the <xyz> man page on your installed system.

2) do not reference the man page alone but a command that will display it, implying that accessing a running system should be the next step, for example "For details, run the 'man <xyz>' (on your system)"


Perhaps there are even more elegant or user-friendly alternatives to achieve the same effect.

Please review this proposal and let me know if it viable for the SSG.

Thank you

mmuehlfeldRH:
>1. mention the installed system explicitly in the reference snippet, for example "For details, see the [xyz] man page on your installed system.
>
>2. do not reference the man page alone but a command that will display it, implying that accessing a running system should be the next step, for example "For details, run the 'man [xyz]' command (on your system)"

In my opinion, both options are not optimal, because we should keep additional resources entries as short as possible. I was told to remove "For details,..." and similar from all my additional resources entries. (I forgot if this was mostly because of the table format of additional resources in Pv2).

I would vote for a variant of option 1:

* `sudoers(5)` man page on your installed system

or even shorter:

* `sudoers(5)` man page on the system

For me it would be clear that I don't run the procedure on one host and then try to open the referenced man page on a different host. WDYT?

My argument against the version with the command is that, in the rest of our modules, commands are (in most cases) shown in text boxes and not inline.

dfitzmau:
Note: Hold off merging this PR until the `On hold` label has been removed. The label in this case flags that a discussion is still in progress.

soulcharmer:
Right, I do not have a strong preference on how exactly the reference snippet should look like, as long as it acknowledges the ambiguity of plainly naming the man page in the docs while not making it readily available on the same platform.

I'd agree that using the "man [xyz]" commands is a bit of a roundabout solution and I'm leaning in the direction of using a simple suffix to indicate the intended location of the man page.

Shall we collect more opinions and/or approvals on this guidance?

bredamc:
Updated the original [issue](https://github.com/redhat-documentation/supplementary-style-guide/issues/224) to suggest adding a general module to explain common info about man pages (details in the issue). And then we could link to that explanatory module whenever we refer to a man page?

Srivaralakshmi:
@bergerhoffer and @bredamc Any progress on this PR?

bergerhoffer:
I've updated the PR to add "on your system" to address the latest comments here.

I also added a sentence mentioning that users run the "man" command to view the information.

The only last thing is just making sure that it is acceptable to have something that is not a link in an additional resources section. @emmurphy1 any updates on this from the mod docs side?

bredamc:
@emmurphy1 @bergerhoffer Ping :) 

@emmurphy1 Can you please confirm whether the mod doc group agreed that it is acceptable to include non-linked text in the Additional resources section?

@bergerhoffer Should we include an example that lists multiple man pages, to show whether to include them on the same line or multiple lines?

mmuehlfeldRH:
> Should we include an example that lists multiple man pages, to show whether to include them on the same line or multiple lines?

+1 for either adding an example or spelling out to list them on the same/multiple lines.


Some additional thoughts from the RHEL admin perspectve:

TL;DR: Man pages and local docs are typically mentioned in our docs for reference purposes (not needed to understand the procedure but they provide additional information, etc.). For this reason, they fit better in additional resources than mentinging x man pages in x different places in the procedure.

Most things on RHEL servers happen on the command line. Tools, config files of services, etc. - they all ship at least one man page (Systemd ships meanwhile 227 and they are great resources!). They all provide a big benefit for admins and they are a great addition to our docs.

In my docs, I explain config options in a way that it's clear what they do, but I often link the corresponding man page in additional resources, because it's a great reference for further details about the options or to figure out other config options that might be helpful for the admin in their own environment. ([Example](https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/9/html/configuring_and_managing_networking/assembly_setting-up-a-wireguard-vpn_configuring-and-managing-networking#proc_configuring-a-wireguard-by-using-nmcli_assembly_setting-up-a-wireguard-vpn))

Many packages also install readme files, example configs, etc. in /usr/share/docs/ ([Example](https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/9/html/deploying_mail_servers/configuring-and-maintaining-a-dovecot-imap-and-pop3-server_deploying-mail-servers#preparing-dovecot-to-use-virtual-users_setting-up-a-dovecot-server-with-pam-authentication)). They are also a great help in many cases and worth referencing them in additional resources - even if they also cannot be linked.

Man pages and local docs are a great help and reference, and they save us from duplicating a lot of information in our guides - at least in RHEL where admins typically work on the command line.

bredamc:
Thanks, Marc! As mentioned in the latest comment in the related Issue 224:
"Discussed at April SSG meeting. Group agreed that we should recommend but not require that man pages be listed on one bullet in a comma-delimited list. @bergerhoffer can you update the PR for this decision?"

bergerhoffer:
Apologies for the delay on updating this. 

Updated to add an example for multiple man pages. I also updated "It is not necessary to link to a website..." to "Do not link to a website...".

Please take another look when you get a chance.

supplementary_style_guide/style_guidelines/formatting.adoc
```diff
@@ -8,6 +8,21 @@ Follow the _IBM Style_ guide advice of _day Month year_ for date formats, for ex
 
 When the format _day Month year_ causes a presentation or clarity issue, use _Month day, year_ (for example, October 3, 2019) instead.
 
+[[man-pages]]
+== Man page references
+
+When referencing a man page in an additional resources section, use the following format: `_<name_of_utility>(<section_number>)_` _man page on your system_.
```
bredamc, line 14: 1. Should we highlight "Additional resources" in some way?
2. Would the text be easier to read if the format was on a new line?

supplementary_style_guide/style_guidelines/formatting.adoc
```diff
@@ -8,6 +8,21 @@ Follow the _IBM Style_ guide advice of _day Month year_ for date formats, for ex
...
+
+Do not link to a website containing the man page information. Users run the `man` command on their system to view the information for the named command or utility.
```
bredamc, line 16: Should we explain that the man-page content can vary from one system to another, and the user must run the `man` command on the system where the named command will be run, to ensure that they see the relevant system-specific information?

supplementary_style_guide/style_guidelines/formatting.adoc
mmuehlfeldRH: 1. Technical nitpick: `<name_of_utility>` is only correct for section 8 and 1. There are also man pages for config files `NetworkManager.conf(5)`, system calls `exit(2)`, libraries `exit(3)`, and others . I would recommend `<man_page_name>(<section_number>)`
2. Is there a special reason for italic on "man page on your system"? The [ISG](https://www.ibm.com/docs/en/ibm-style?topic=references-citations#general-guidelines) does not use italic for qualifiers (`For more information, see the _DQTI_ book.`)

supplementary_style_guide/style_guidelines/formatting.adoc
bredamc: I meant that I think it's useful to explain _why_ we're including the text "on your system" in the examples.

mmuehlfeldRH: If it's about adding the why to SSG (not to the actual entries), I agree that it would be helpful.

mportman12:
@bergerhoffer PTAL at outstanding review comments.

supplementary_style_guide/style_guidelines/formatting.adoc
libander: If we're adding the why to the SSG, it is important to note that the content of man files is determined not by OS version, but by kernel version. You can have multiple kernel versions across the life of an OS Version.

supplementary_style_guide/style_guidelines/formatting.adoc
mmuehlfeldRH: @libander I don't understand your comment. How is the `man` man page content you posted relevant for the proposed change in the SSG?

supplementary_style_guide/style_guidelines/formatting.adoc
mmuehlfeldRH: @libander what do you mean with "the content of man files is determined not by OS version, but by kernel version"?

Man pages are shipped with the RPMs of tools, services, etc. they belong to. For example, `find(1)` is provided by the `findutils` RPM, and `rsync(1)` by the `rsync` package. If you update one of these packages, then the man pages are also updated (but the man page content only changes if the new RPM also contains a different man page). Anyway, man pages don't change if you boot a different kernel.

None of the kernel packages (`kernel`, `kernel-core`, `kernel-modules`, `kernel-modules-extra`, `kernel-modules-core`) contains a man page. I don't know if there are other packages than `kernel*` that can be installed at the same time in different versions, but if such packages contain man pages, then the old man pages are replaced by the new ones because the files use the same paths to the man page files. If you check, for example, `/usr/share/man/man1/` you see that there's only one man page for each tool.

In case that you meant something else, can you please explain it further?

libander: > In case that you meant something else, can you please explain it further?

You're right, many packages come with their own man-page included. The [`man-pages.rpm`](https://access.redhat.com/downloads/content/rhel---9/x86_64/11051/man-pages/6.04-1.el9/src/fd431d51/package) provides man-pages for [GNU/Linux](https://git.kernel.org/pub/scm/docs/man-pages/man-pages.git/about/) system and library calls, etc. 

> adding the why to SSG (not to the actual entries)

It was with this context in mind I was suggesting that we be slightly more specific than 'vary by system'  -- 'vary by package or kernel version' would be less mysterious. :)

supplementary_style_guide/style_guidelines/formatting.adoc
libander: > 1. Technical nitpick: `<name_of_utility>` is only correct for section 8 and 1. There are also man pages for config files `NetworkManager.conf(5)`, system calls `exit(2)`, libraries `exit(3)`, and others . I would recommend `<man_page_name>(<section_number>)`

Likewise, given that the man pages themselves classify the pages: 
> Each page argument given to man is normally the name of a program, utility or function.

I'd consider changing line 16's: 
`to view the information for the named command or utility.` 
to: 
`to view the information for the program, utility or function.`

mportman12:
@bergerhoffer Please take a look at outstanding review comments.

bergerhoffer:
Updated per the latest feedback. PTAL: https://file.rdu.redhat.com/~ahoffer/2024/main-man-pages.html#man-pages

bredamc:
Many thanks for the preview link, @bergerhoffer! Very helpful :)

Not sure if you didn't see my comments of 27 Sept or are choosing to not implement them, so I'll reiterate them here, just in case :)

1.     Should we highlight "Additional resources" in some way?
2.     Would the text be easier to read if the format was on a new line?

I'll also add that I think we should not use italics for the text "man page on your system", because that text is to be reproduced verbatim.

supplementary_style_guide/style_guidelines/formatting.adoc
```diff
@@ -8,6 +8,21 @@ Follow the _IBM Style_ guide advice of _day Month year_ for date formats, for ex
 
 When the format _day Month year_ causes a presentation or clarity issue, use _Month day, year_ (for example, October 3, 2019) instead.
 
+[[man-pages]]
+== Man page references
+
+When referencing a man page in an additional resources section, use the following format: `_<man_page_name>(<section_number>)_` _man page on your system_.
```
bredamc, line 14:
```suggestion
When referencing a man page in an "Additional resources" section, use the following format: 
* `_<man_page_name>(<section_number>)_` man page on your system
```

supplementary_style_guide/style_guidelines/formatting.adoc
```diff
@@ -8,6 +8,21 @@ Follow the _IBM Style_ guide advice of _day Month year_ for date formats, for ex
...
+
+Do not link to a website containing the man page information. The contents of a man page may vary between systems or package versions, so users must run the `man` command on their system to view the most accurate information for the named command or utility.
```
bredamc, line 16:
```suggestion
Do not link to a website that contains the man page information. 
The contents of a man page might vary between systems or package versions, so users must run the `man` command on the target system to view the system-specific information for the named command or utility.
```

bredamc (commented):
A few minor comments.

supplementary_style_guide/style_guidelines/formatting.adoc
bergerhoffer: No sorry, I missed this suggestion! I agree that this looks better, will update.

supplementary_style_guide/style_guidelines/formatting.adoc
bergerhoffer: Thanks! I used your wording suggestions, but kept them on the same line (because the info about users having to run the `man` command on their system is a reason for why we say not to link to man page websites)

bergerhoffer:
@bredamc sorry for missing your suggestions earlier! Updated if you can please take another look

supplementary_style_guide/style_guidelines/formatting.adoc
bredamc: Putting them on separate lines was just to make it easier to review (each sentence on a separate line) -- it didn't affect the output, the two sentences were still in the same paragraph. But it was a tiny suggestion, no problem at all for not implementing it :)

supplementary_style_guide/style_guidelines/formatting.adoc
```diff
@@ -8,6 +8,23 @@ Follow the _IBM Style_ guide advice of _day Month year_ for date formats, for ex
 
 When the format _day Month year_ causes a presentation or clarity issue, use _Month day, year_ (for example, October 3, 2019) instead.
 
+[[man-pages]]
+== Man page references
+
+When referencing a man page in an "Additional resources" section, use the following format:
+
+* `_<man_page_name>(<section_number>)_` _man page on your system_
```
bredamc, line 16:
```suggestion
* `_<man_page_name>_(_<section_number>_)` man page on your system
```
The text "man page on your system" should not be in italics.  It's confusing when it's in italics, because it seems as if it should be replaced by something else, when in fact it should be reproduced verbatim.

The parentheses shouldn't be in italics either, but that's not as bothersome :)

supplementary_style_guide/style_guidelines/formatting.adoc
bergerhoffer: Gotcha - yeah I normally don't write my sentences that way like I know some people do :)

supplementary_style_guide/style_guidelines/formatting.adoc
bergerhoffer: Thanks, updated!

bergerhoffer:
@bredamc Updated, ready for another look! Didn't update the preview yet though because something is erroring out when I try the transfer

Srivaralakshmi:
@bredamc Can you please take another look at this revised PR? TY!

bredamc (approved):
LGTM -- thanks, @bergerhoffer!
(And thanks for the reminder to approve, @Srivaralakshmi!)

Srivaralakshmi:
@bburt-rh , @mportman12 , @ncbaratta Need two more approvals. Please help. TY!

mportman12 (approved):
LGTM

Srivaralakshmi:
Thank you all!

PR #298, opened 2023-02-28
Add new GitHub actions publishing workflow, removes .travis.yml

aireilly:
Ensure that the GitHub Pages site is published from gh-pages branch.

Preview: https://aireilly.github.io/supplementary-style-guide/

image


.github/workflows/deploy-site.yml
```diff
@@ -0,0 +1,31 @@
...
+      - name: Install AsciiDoctor
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y asciidoctor
+
+      - name: Build AsciiDoc
+        run: |
+          asciidoctor supplementary_style_guide/main.adoc -o index.html
+          mkdir docs
+          mv index.html docs/
```
bergerhoffer, line 25: We'll need to add something here to copy the images/ folder too

bergerhoffer (commented):
@aireilly you are awesome, thank you so much!

I'll take a closer look through, but this is the first thing I noticed

bergerhoffer (dismissed):
I won't really know til we merge it and test it, but this LGTM for merging. Thanks again!

mportman12:
Is there a way to do a staged merge without taking it to production?

.github/workflows/deploy-site.yml
```diff
@@ -0,0 +1,32 @@
...
+        run: |
+          asciidoctor supplementary_style_guide/main.adoc -o index.html
+          mkdir docs
+          cp -r supplementary_style_guide/images/ docs/images
+          mv index.html docs/
+
+      - name: Publish to GitHub Pages
+        uses: peaceiris/actions-gh-pages@v3
+        with:
+          github_token: ${{ secrets.GITHUB_TOKEN }}
```
bergerhoffer, line 31: @aireilly is this GITHUB_TOKEN something we have to configure somewhere?

bburt-rh:
Is there a reason why the `[discrete]` code is being added to `s.adoc`? Otherwise this LGTM, though I also have the question mentioned by @bergerhoffer re the GH token.

mportman12:
> Is there a reason why the `[discrete]` code is being added to `s.adoc`? Otherwise this LGTM, though I also have the question mentioned by @bergerhoffer re the GH token.

@bburt-rh I forgot to add [discrete] when I added the sticky glossary entries, and it didn't get caught in review.

bergerhoffer:
> Is there a way to do a staged merge without taking it to production?

@mportman12 Aidan did share this with me, his test for his own fork: https://aireilly.github.io/supplementary-style-guide/

So that was good enough for me. Plus we _hopefully_ can't mess this up too bad since it's not updating at all currently. Though I will go ahead and make a backup of the gh-pages branch right now just in case - so we can restore if that gets messed up during this testing.

bburt-rh (dismissed):

bburt-rh:
Approved

.github/workflows/deploy-site.yml
aireilly: Should be available by default.

bergerhoffer: FYI for anyone else curious, it's this: https://docs.github.com/en/actions/security-guides/automatic-token-authentication

bergerhoffer:
Unless there are any objections, will plan to merge this tomorrow morning (eastern time) and see how it goes! :crossed_fingers:

dfitzmau (dismissed):

bredamc (approved):
Looks good enough to try a test! (Caveat that I don't fully understand the code.)

bergerhoffer:
Merging to test and see how this goes!

PR #310, opened 2023-04-19
Update user-replaced value guidance for output

sbmetz:
Addresses issue 303, add guidance for replaceable variables in output. In "User-replaced values" section, added a bullet to describe the council's recommendation, and added an example. Should we also make parallel changes for the "User-replaced values for XML" section?

### DIFF
--- a/supplementary_style_guide/style_guidelines/formatting.adoc
+++ b/supplementary_style_guide/style_guidelines/formatting.adoc
@@ -146,12 +146,13 @@ $ oc describe node __<node_name>__
 
 .Example AsciiDoc: User-replaced value in an output example
 
-----
-connection.id:              __<profile_name>__
-connection.uuid:            b6cdfa1c-e4ad-46e5-af8b-a75f06b79f76
-connection.type:            802-3-ethernet
-connection.interface-name:  enp7s0
-----
+  [subs="+quotes"]
+  ----
+  connection.id:              __<profile_name>__
+  connection.uuid:            b6cdfa1c-e4ad-46e5-af8b-a75f06b79f76
+  connection.type:            802-3-ethernet
+  connection.interface-name:  enp7s0
+  ----
 
 This renders as:
 
--- a/supplementary_style_guide/style_guidelines/formatting.adoc
+++ b/supplementary_style_guide/style_guidelines/formatting.adoc
@@ -112,6 +112,7 @@ Ensure that user-replaced values have the following characteristics:
 * Lowercase, unless the rest of the related text is uppercase or another capitalization scheme
 * Italicized
 * If the user-replaced value is referencing a value in code or in a command that is normally monospace, also use monospace for the user-replaced value
+* If the user-replaced value appears in example output, only format the value in the output if the replaceable value was also in the command or if formatting adds value to the customer. Otherwise, leave the replaceable value unformatted as an example.
 
 .Example AsciiDoc: User-replaced value in a paragraph
 
@@ -140,6 +141,29 @@ This renders as:
 [subs="+quotes"]
 ----
 $ oc describe node __<node_name>__
+----
+====
+
+.Example AsciiDoc: User-replaced value in an output example
+
+----
+# *nmcli connection show __<profile_name>__*
+connection.id:              __<profile_name>__
+connection.uuid:            b6cdfa1c-e4ad-46e5-af8b-a75f06b79f76
+connection.type:            802-3-ethernet
+connection.interface-name:  enp7s0
+----
+
+This renders as:
+
+====
+[subs="+quotes"]
+----
+# nmcli connection show __<profile_name>__
+connection.id:              __<profile_name>__
+connection.uuid:            b6cdfa1c-e4ad-46e5-af8b-a75f06b79f76
+connection.type:            802-3-ethernet
+connection.interface-name:  enp7s0
 ----
 ====
 
--- a/supplementary_style_guide/style_guidelines/formatting.adoc
+++ b/supplementary_style_guide/style_guidelines/formatting.adoc
@@ -147,7 +147,6 @@ $ oc describe node __<node_name>__
 .Example AsciiDoc: User-replaced value in an output example
 
 ----
-# *nmcli connection show __<profile_name>__*
 connection.id:              __<profile_name>__
 connection.uuid:            b6cdfa1c-e4ad-46e5-af8b-a75f06b79f76
 connection.type:            802-3-ethernet
@@ -159,7 +158,6 @@ This renders as:
 ====
 [subs="+quotes"]
 ----
-# nmcli connection show __<profile_name>__
 connection.id:              __<profile_name>__
 connection.uuid:            b6cdfa1c-e4ad-46e5-af8b-a75f06b79f76
 connection.type:            802-3-ethernet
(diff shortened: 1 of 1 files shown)


supplementary_style_guide/style_guidelines/formatting.adoc
```diff
@@ -143,6 +144,29 @@ $ oc describe node __<node_name>__
...
+connection.type:            802-3-ethernet
+connection.interface-name:  enp7s0
+----
+
+This renders as:
+
+====
+[subs="+quotes"]
+----
+# nmcli connection show __<profile_name>__
```
bergerhoffer, line 162: On OpenShift, we have a rule where we separate commands from output. This helps with being able to use the "copy" button to copy the command.

This is the first time we're introducing this (command+output) to the SSG as far as I'm aware, so I don't want to introduce it this way if we really should be separating them. 

Do others separate command from output, or keep them in the same block?

supplementary_style_guide/style_guidelines/formatting.adoc
```diff
@@ -112,6 +112,7 @@ Ensure that user-replaced values have the following characteristics:
 * Lowercase, unless the rest of the related text is uppercase or another capitalization scheme
 * Italicized
 * If the user-replaced value is referencing a value in code or in a command that is normally monospace, also use monospace for the user-replaced value
+* If the user-replaced value appears in example output, only format the value in the output if the replaceable value was also in the command or if formatting adds value to the customer. Otherwise, leave the replaceable value unformatted as an example.
```
bergerhoffer, line 115: Right now this reads to me like the options are to format it or not format it. But I think if you're using a replaceable value in output, it still needs formatted properly.

Maybe we could reword to be clearer that the options are to either use a replaceable value (and italicize / put it in angle brackets) or use an example value (and don't format it any special way).

```suggestion
* If you want to use a user-replaced value in example output, format the replaceable value with italics and in angle brackets. Alternatively, if you choose to use an example value instead, do not italicize the example value or place it in angle brackets.
```

bburt-rh: +1 to the suggested edit from @bergerhoffer. They distinguish the two alternatives more clearly.

bburt-rh:
I'm good with this PR as soon as the comments from @bergerhoffer are addressed.

sbmetz:
I accepted/commited @bergerhoffer 's first suggestion. 
@redhat-documentation/ccs-style-council -- any other thoughts on separating command/output?

supplementary_style_guide/style_guidelines/formatting.adoc
mmuehlfeldRH: I like SSG enhancement as currently proposed by this pull request.

Using separate blocks for command and output would require an additional intro sentence for a output block, visually increases the length of a procedure, and doesn't look nice (just my opinion).

In RHEL, most things happen on the command line, and we have commands in almost every step of a procedure. Sure, in most cases we only show the command without output, but separating the command from the output in existing docs would still be a lot of extra work.

I understand the argument of copy/paste, but I think this should be solved by JavaScript on the Customer Portal. For example, only the bold text (command) should be copied to the clipboard, when you click the copy button.

bergerhoffer: We actually don't require another sentence introducing it. We use the heading of ".Example output", and include it right after the code block. 

An example of that is here: https://docs.openshift.com/container-platform/4.13/security/audit-log-view.html#nodes-nodes-audit-log-basic-viewing_audit-log-view

Or how it looks on the portal: https://access.redhat.com/documentation/en-us/openshift_container_platform/4.13/html/security_and_compliance/audit-log-view#nodes-nodes-audit-log-basic-viewing_audit-log-view

It could be just because I'm used to it, but I prefer it this way, both because of the ability to use the "Copy" button and just get the command, but also visually so you can distinguish the command from the output. I'm seeing here that the command looks like it's trying to be bolded, which would help visually distinguish, but again I'm not sure if that's what people normally do.

But really, I'm not trying to debate which method is better or which we should be doing. Because that's not what we're trying to explicitly add guidance for here. But we are going to be implicitly adding guidance for however we do this. If we add it in here in the same block, and bold the command, then people are going to take that as guidance for how they should be doing it.

(Also FYI @sbmetz that it doesn't render as bold, if that was your intent. You'll need to pass in subs to make that render properly.)

I'm just wondering if there's a way we can do it that avoids this issue all together. Could we just show the output perhaps w/o the command? That's really the only thing we're trying to convey here. Thoughts?

IngridT1: I think that we should show output without the command, because RHEL and OpenShift are doing this so differently.

joaedwar (dismissed):
lgtm

joaedwar:
Hi @sbmetz. I approved this PR;it looks good to me!

@bburt-rh @bergerhoffer @mportman12 any additional comments? Is this okay to merge?

mportman12:
@joaedwar We wait to merge until there are three approvals, so not yet.

supplementary_style_guide/style_guidelines/formatting.adoc
```diff
@@ -112,6 +112,7 @@ Ensure that user-replaced values have the following characteristics:
 * Lowercase, unless the rest of the related text is uppercase or another capitalization scheme
 * Italicized
 * If the user-replaced value is referencing a value in code or in a command that is normally monospace, also use monospace for the user-replaced value
+* If you want to use a user-replaced value in example output, format the replaceable value with italics and in angle brackets. Alternatively, if you choose to use an example value instead, do not italicize the example value or place it in angle brackets.
```
bburt-rh, line 115:
```suggestion
* If you want to use a user-replaced value in example output, format the replaceable value with italics and in angle brackets. Alternatively, if you choose to use an example value instead, do not italicize the example value or place it in angle brackets.
```
That last sentence is a bit ambiguous to my ears. Not sure if it means "do not italicize the example value **and** do not place it in angle brackets" or does it mean "do not italicize the example value. Or, alternatively, place it in angle brackets."

bburt-rh:
@sbmetz I added one comment/suggestion. I'd be interested to hear your take.

supplementary_style_guide/style_guidelines/formatting.adoc
sbmetz: Thanks, Brian -- I take your point about ambiguity. I'm pretty sure I meant `do not italicize the example value and do not place it in angle brackets` and your suggestion looks good to me.

joaedwar (dismissed):
lgtm

supplementary_style_guide/style_guidelines/formatting.adoc
```diff
@@ -143,6 +144,27 @@ $ oc describe node __<node_name>__
 ----
 ====
 
+.Example AsciiDoc: User-replaced value in an output example
+
```
bergerhoffer, line 148: I think you should add the `[subs="+quotes"]` line above here, so people know it's required.

Also then if you indent all of the lines associated with this code block, it will show them all as is. See lines 133-136 for how the above example does it

bergerhoffer (commented):
One last adjustment to make it look similar to the above example

supplementary_style_guide/style_guidelines/formatting.adoc
sbmetz: OK, the example starting at ~line 147 now matches the ~line 131 example.

bergerhoffer: LGTM, thanks for the update!

bergerhoffer (approved):
LGTM!

joaedwar (approved):
lgtm

Issue #347, opened 2023-06-25
Clarify markup for UI elements that do not have a tooltip or label


apinnick:
Section to update: [UI Elements](https://redhat-documentation.github.io/supplementary-style-guide/#user-interface-elements)

> Bold all graphical user interface elements, including, but not limited to, menu names, menu items, button names, dialog box names, and window names. Bold the item, even if it is not clickable.

I recommend adding a sentence to clarify markup and capitalization for UI elements that do not have a tooltip or label. The current text could mean that a search field, for example, should be bolded. Same applies to icons.

Which is correct?
- **Search** field
- **search** field
- search field

bburt-rh:
I agree with @bergerhoffer that if the actual text doesn't appear in the UI then it should not be bolded or capitalized (or capitalized only if referring to a proper noun).

The current IBM style guidance for UI fields does not cover cases where there is no label or tooltip for the field: https://www.ibm.com/docs/en/ibm-style?topic=elements-ui#field.

I would be in favor of adding guidance for this use case to https://redhat-documentation.github.io/supplementary-style-guide/#user-interface-elements: something like, "when no label, tooltip, or other identifying text exists for a field, make a generic reference to it. For example, if a search field contains no label or tool tip, refer to it as `the search field`."

bredamc:
I think this item is already covered in the _UI elements_ topic in _IBM Style_. See the _icon_ entry in the usage table:

If an icon opens a menu, refer to the icon by its label, followed by icon. 
Example: 
_For an icon with the label **Actions** that opens a menu:_
Click the **Actions** icon.

If an icon does not have a label or tooltip, refer to the icon by a generic description.
Example: 
Click the print icon.

bergerhoffer:
@bredamc thanks for sharing that. Though since we do override the ISG's guidance on how to style UI elements (we just bold everything instead of getting particular depending on the element), I think we should also mention this in our section in the SSG? Does that sound okay to you?

bredamc:
Ah my apologies, I misunderstood the intention! Yes, absolutely -- thanks, @bergerhoffer!

mportman12:
@bergerhoffer makes a good point about our UI guidance. I think this deserves a sentence to note that we align with the ISG on this.

bergerhoffer:
I can work on a draft PR for this

bergerhoffer:
All - I have a PR up, please take a look and let me know what feedback you have. Thanks!

[...]

PR #349, opened 2023-06-27
GitHub-347: Clarifying not to bold unlabeled GUI elements

bergerhoffer:
To address #347

Preview: https://file.rdu.redhat.com/~ahoffer/2023/main-ui-elements.html#user-interface-elements
(must be on VPN + ignore that the style looks different)

### DIFF
--- a/supplementary_style_guide/style_guidelines/graphical-interfaces.adoc
+++ b/supplementary_style_guide/style_guidelines/graphical-interfaces.adoc
@@ -2,7 +2,7 @@
 [[graphical-interfaces]]
 = Graphical interfaces
 
-For more detailed guidance on how to document UI elements, see link:https://www.patternfly.org/v4/ux-writing/about[PatternFly].
+For more detailed guidance on how to document user interface (UI) elements, see link:https://www.patternfly.org/v4/ux-writing/about[PatternFly].
 
 [[microcopy]]
 == Microcopy
@@ -29,12 +29,16 @@ In the *Name* field, enter `test-postgresql`.
 [[user-interface-elements]]
 == User interface elements
 
-Bold all graphical user interface elements, including, but not limited to, menu names, menu items, button names, dialog box names, and window names. Bold the item, even if it is not clickable.
+Use bold text for all graphical user interface (GUI) element names, including menus, menu items, buttons, dialog boxes, and windows. Use bold text for the element name if the name appears in the GUI, even if the element is not clickable.
 
 .Example AsciiDoc
 ----
 On the *Installed Operators* page, click *Metering*.
 ----
+
+If an element is not labeled in the GUI, refer to the element by a generic description and do not use bold text. For example, if a search field is not labeled in the GUI, write it as "the search field", not "the *Search* field".
+
+
 
 
 // TODO: Add new style entries alphabetically in this file


supplementary_style_guide/style_guidelines/graphical-interfaces.adoc
```diff
@@ -29,7 +29,9 @@ In the *Name* field, enter `test-postgresql`.
 [[user-interface-elements]]
 == User interface elements
 
-Bold all graphical user interface elements, including, but not limited to, menu names, menu items, button names, dialog box names, and window names. Bold the item, even if it is not clickable.
+Bold all graphical user interface (GUI) element names, including, but not limited to, menu, menu item, button, dialog box, and window names. Bold the element name if it appears in the GUI, even if it is not clickable.
+
+If the element is not labeled in the GUI, refer to it by a generic description and do not bold it. For example, if a search field is not labeled in the GUI, write it as "the search field", not "the *Search* field".
```
bredamc, line 34:
```suggestion
If an element is not labeled in the GUI, refer to the element by a generic description and do not use bold text. For example, if a search field is not labeled in the GUI, write it as "the search field", not "the *Search* field".
```

supplementary_style_guide/style_guidelines/graphical-interfaces.adoc
```diff
@@ -29,7 +29,9 @@ In the *Name* field, enter `test-postgresql`.
 [[user-interface-elements]]
 == User interface elements
 
-Bold all graphical user interface elements, including, but not limited to, menu names, menu items, button names, dialog box names, and window names. Bold the item, even if it is not clickable.
+Bold all graphical user interface (GUI) element names, including, but not limited to, menu, menu item, button, dialog box, and window names. Bold the element name if it appears in the GUI, even if it is not clickable.
```
bredamc, line 32:
```suggestion
Use bold text for all graphical user interface (GUI) element names, including, but not limited to, menu, menu item, button, dialog box, and window names. Use bold text for the element name if the name appears in the GUI, even if the element is not clickable.
```

A few minor comments:

1. Bold is an adjective, not a verb. :) 
2. Some of the pronouns were ambiguous -- please check my suggested rewrites, because I might have misinterpreted the original pronoun targets!

bredamc (commented):
Just a few minor comments for your consideration :)

supplementary_style_guide/style_guidelines/graphical-interfaces.adoc
bergerhoffer: I thought about "bold" as a verb - went as far as checking merriam webster :smile: But I couldn't think of a better alternative and figured I'd leave it alone since it was already there. Thanks for making a great suggestion!

bergerhoffer:
Thanks @bredamc, updated per your feedback!

bburt-rh (approved):

bburt-rh:
lgtm

supplementary_style_guide/style_guidelines/graphical-interfaces.adoc
```diff
@@ -29,7 +29,9 @@ In the *Name* field, enter `test-postgresql`.
 [[user-interface-elements]]
 == User interface elements
 
-Bold all graphical user interface elements, including, but not limited to, menu names, menu items, button names, dialog box names, and window names. Bold the item, even if it is not clickable.
+Use bold text for all graphical user interface (GUI) element names, including, but not limited to, menu, menu item, button, dialog box, and window names. Use bold text for the element name if the name appears in the GUI, even if the element is not clickable.
+
+If an element is not labeled in the GUI, refer to the element by a generic description and do not use bold text. For example, if a search field is not labeled in the GUI, write it as "the search field", not "the *Search* field".
```
dfitzmau, line 34: I wonder if adding a note would be useful (I know the PatternFly link has already been provided) to inform a doc user that if they do not know the name of a tooltip item then they should refer to the Components section of the PatternFly docs? The kebab icon comes to mind, but luckily this is not under the *Components* section list, but the term "kebab" is embedded in the Dropdown sub-section. 

https://www.patternfly.org/v4/components/overflow-menu/html
https://www.patternfly.org/v4/components/dropdown/html/#kebab

bredamc: Might be safer not to include such links? The term "kebab icon" should not be used in our docs :)

dfitzmau: Yes, best not to link to the "dropdown menu" link, but the general "components" link would be useful. We already mention PatternFly, so one could locate the kebab menu if they go digging.

supplementary_style_guide/style_guidelines/graphical-interfaces.adoc
```diff
@@ -29,7 +29,9 @@ In the *Name* field, enter `test-postgresql`.
...
 
 .Example AsciiDoc
```
apinnick, line 36: You might want to move the Example so that it appears after the first paragraph (Use bold text...)

If you want to add an example to the new paragraph, you could use "Click the edit icon." or "Enter the term in the search field."

apinnick, line 36: If this example is to show bolding GUI elements, it should be moved to a position before your new section.

bergerhoffer: Good suggestion, done!

supplementary_style_guide/style_guidelines/graphical-interfaces.adoc
bergerhoffer: I think I'd prefer not to link to PatternFly here, because the ISG does override some of the guidance there with the names of the elements. So it's probably the ISG that people should refer to more closely for if they don't know what the UI element is called. (And we can't link to the ISG).

supplementary_style_guide/style_guidelines/graphical-interfaces.adoc
```diff
@@ -29,12 +29,16 @@ In the *Name* field, enter `test-postgresql`.
 [[user-interface-elements]]
 == User interface elements
 
-Bold all graphical user interface elements, including, but not limited to, menu names, menu items, button names, dialog box names, and window names. Bold the item, even if it is not clickable.
+Use bold text for all graphical user interface (GUI) element names, including, but not limited to, menu, menu item, button, dialog box, and window names. Use bold text for the element name if the name appears in the GUI, even if the element is not clickable.
```
mportman12, line 32:
```suggestion
Use bold text for all graphical user interface (GUI) element names, including menu, menu item, button, dialog box, and window names. Use bold text for the element name if the name appears in the GUI, even if the element is not clickable.
```
Redundant and gets rid of an extra layer of commas.

mportman12 (changes requested):
One small suggestions, otherwise LGTM.

supplementary_style_guide/style_guidelines/graphical-interfaces.adoc
bergerhoffer: Updated. Also made the items plural to keep it simpler and read a little better (I think at least). Let me know what you think!

dfitzmau (approved):

bburt-rh (approved):

mportman12 (approved):
LGTM

Issue #334, opened 2023-05-31
Update the SSG's IP address glossary entry


tmalove:
Description:
This issue is from the SSG survey feedback from May 2023 for the question - What about the guide could be improved? What changes would you like to see?

Feedback:
The references for IP addresses reserved for documentation are RFC-5737 and RFC 3849, not the ISG. They are under IANA's authority, not IBM's, and in the event of disagreement, IBM is wrong.

Action required:
Verify/research if the 'IP addresses' glossary entry needs to be updated.

bergerhoffer:
The feedback is correct that it is under IANA's authority. Though the IBM style guide does link to the RFC 5737 (for IPv4) and RFC 3849 (for IPv6).

I'm torn on whether we need to make an update or not. We could make an update to link only to the RFCs [1] [2] and actually put the values in the SSG, like we do for the mac addresses section. But I think that the IBM section on IP addresses does actually provide some additional beneficial information.

Or we could just make a minor adjustment to the wording and say that they are defined by RFCs (and provide links) and to see the section in the ISG for more info.

Curious what others think here.

[1] https://www.rfc-editor.org/rfc/rfc5737.html
[2] https://www.rfc-editor.org/rfc/rfc3849.html

bburt-rh:
@bergerhoffer Where does the PR for this issue stand at the moment? Any updates? Ready for review?

PR #358, opened 2023-07-06
GitHub-334: Adding details and RFCs for IP address ranges

bergerhoffer:
Expanded the IP addresses section to point out that the guidance on ranges to use actually comes from the RFCs.

Fixes: #334

Preview: https://file.rdu.redhat.com/~ahoffer/2023/main-ip-address.html#ip-addresses-and-mac-addresses
(VPN required; ignore old style)

### DIFF
--- a/supplementary_style_guide/style_guidelines/code-commands.adoc
+++ b/supplementary_style_guide/style_guidelines/code-commands.adoc
@@ -29,14 +29,56 @@ If you include a shell prompt to indicate that a user with root privileges must 
 [[ip-addresses-and-mac-addresses]]
 == IP addresses and MAC addresses
 
-Use the IP and MAC address ranges that are reserved for documentation purposes to avoid the likelihood of conflicts and confusion:
+Use the IP and MAC address ranges that are reserved for documentation purposes to avoid the likelihood of conflicts and confusion.
 
-* Reserved IP address ranges: See the _IBM Style_ guide for the list of reserved IP address ranges to use in documentation.
+[[reserved-ip-addresses]]
+=== Reserved IP addresses
 
-* Reserved MAC addresses (defined in link:https://www.rfc-editor.org/rfc/rfc7042.html#section-2.1.2[RFC 7042]):
+Reserved IPv4 addresses for documentation are defined in link:https://www.rfc-editor.org/rfc/rfc5737.html[RFC 5737]:
 
-** For unicast: 00:00:5E:00:53:00 - 00:00:5E:00:53:FF
-** For multicast: 01:00:5E:90:10:00 - 01:00:5E:90:10:FF
+[cols="1,1,1", options="header"]
+|===
+|Network
+|Subnet mask
+|Assignable addresses
+
+|192.0.2.0
+|24
+|192.0.2.1 - 192.0.2.254
+
+|198.51.100.0
+|24
+|198.51.100.1 - 198.51.100.254
+
+|203.0.113.0
+|24
+|203.0.113.1 - 203.0.113.254
+|===
+
+Reserved IPv6 addresses for documentation are defined in link:https://www.rfc-editor.org/rfc/rfc3849.html[RFC 3849]:
+
+[cols="1,1,1", options="header"]
+|===
+|Network
+|Prefix
+|Assignable addresses
+
+|2001:0DB8::
+(long form: 2001:0DB8:0:0:0:0:0:0)
+|32
+|2001:0db8:0000:0000:0000:0000:0000:0000 - 2001:0db8:ffff:ffff:ffff:ffff:ffff:ffff
+
+|===
+
+See the _IBM Style_ guide for additional guidance on using IP addresses in documentation.
+
+[[reserved-mac-addresses]]
+=== Reserved MAC addresses
+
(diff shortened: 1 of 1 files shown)


supplementary_style_guide/style_guidelines/code-commands.adoc
```diff
@@ -29,14 +29,30 @@ If you include a shell prompt to indicate that a user with root privileges must
...
-* Reserved IP address ranges: See the _IBM Style_ guide for the list of reserved IP address ranges to use in documentation.
+[[reserved-ip-addresses]]
+=== Reserved IP addresses
 
-* Reserved MAC addresses (defined in link:https://www.rfc-editor.org/rfc/rfc7042.html#section-2.1.2[RFC 7042]):
+Reserved IPv4 addresses for documentation are defined in link:https://www.rfc-editor.org/rfc/rfc5737.html[RFC 5737]:
 
-** For unicast: 00:00:5E:00:53:00 - 00:00:5E:00:53:FF
-** For multicast: 01:00:5E:90:10:00 - 01:00:5E:90:10:FF
+* 192.0.2.0/24 (or 192.0.2.*)
```
dfitzmau, line 39: From looking at the RFC, I do not directly see information pertaining to using blocks other than 192.0.2.0/24, 198.51.100.0/24, and 203.0.113.0/24. The asterisk might indicated that one can use the address on any port, which might be OK?

dfitzmau:
Thanks for updating these entries, @bergerhoffer . I added one comment inline. I think I'm missing something, so apologize if my comment is inaccurate.

supplementary_style_guide/style_guidelines/code-commands.adoc
bergerhoffer: @dfitzmau I admit that i am no expert on IP addresses. I took these values from what the IBM style guide listed here: https://www.ibm.com/docs/en/ibm-style?topic=elements-web-ip-email-addresses#using-addresses-in-examples

@mmuehlfeldRH I know that you submitted the section on mac addresses. Are you familiar enough with IP addresses too to give some feedback on what we should list for IP addresses? I know that you gave an actual range for mac addresses

mmuehlfeldRH: > The asterisk might indicated that one can use the address on any port, which might be OK?

@dfitzmau In the network world, a "port" identifies an endpoint that a service uses (for example, HTTPS uses, by default, port 443). Services can listen on a specific port on a specific IP address in a host (e.g. 192.0.2.1:443), but this has nothing to do with IP addresses.


> Are you familiar enough with IP addresses too to give some feedback on what we should list for IP addresses?

@bergerhoffer If you want to show a range instead of 192.0.2.* (which is not a typical notation), then we can show: 192.0.2.0 - 192.0.2.255

Some additional information:

A /24 subnet mask in an IPv4 network means that the last octet ("block") of the address can be anything between 0 and 255, but the 0 indicates in this case the network itself and 255 is the broadcast address.

IPv4 example: In a 192.0.2.0/24 network:

- You can assign addresses from the range 192.0.2.1 - 192.0.2.254 to your hosts.
- 192.0.2.0 stands for the network itself. You must not assign this address to any device.
- 192.0.2.254 is the broadcast address in this network. You must not assign this address to any device.

It works the same for all IPv4 /24 networks.

bergerhoffer:
Again, apologies on the delay in updating this one. 

@mmuehlfeldRH I updated the IPv4 addresses to be ranges if you can make sure it looks right now. Also, should we do a range for the IPv6 one (so that we have ranges for all of them)? I'm not sure how that one would look.

mmuehlfeldRH:
@bergerhoffer after thinking about it again, I have two suggestions:

1. We should preserve the subnet masks and prefix information in the descriptions. If we show only ranges, this information not very obvious - especially not for IPv6.

2. It would help writers who are not familiar with IP networks to show which addresses they can assign to devices in their procedures. In IPv4 ranges, the first IP address indicates the network and the last one is the broadcast address and both addresses cannot be assigned to devices.

What do you think about:


Reserved IPv4 addresses for documentation are defined in [RFC 5737](https://www.rfc-editor.org/rfc/rfc5737.html):

| Network      | Subnet mask | Assignable addresses              |
| ------------ | ----------- | ----------------------------- |
| 192.0.2.0    |          24 | 192.0.2.1 - 192.0.2.254       |
| 198.51.100.0 |          24 | 198.51.100.1 - 198.51.100.254 |
| 203.0.113.0  |          24 | 203.0.113.1 - 203.0.113.254   |

Reserved IPv6 addresses for documentation are defined in [RFC 3849](https://www.rfc-editor.org/rfc/rfc3849.html):

<table>
 <tr>
  <th>Network</th><th>Prefix</th><th>Assignable addresses</th>
 </tr>
  <tr>
    <td rowspan="2">2001:0DB8::<br/>(long form: 2001:0DB8:0:0:0:0:0:0)</td>
    <td>32</td>
    <td>2001:0db8:0000:0000:0000:0000:0000:0000 - 2001:0db8:ffff:ffff:ffff:ffff:ffff:ffff</td>
  </tr>
</table>

mmuehlfeldRH:
@bergerhoffer LGTM. Thanks for the update.

dfitzmau (approved):
Links work, preview renders. LGTM!

lahinson (approved):
lgtm

Issue #366, opened 2023-07-25
Ingress definition


skrthomas:
Should we clarify that ingress can also refer to incoming traffic (like egress and ingress traffic), and in that case its lowercase, as opposed to Ingress resources, which are capitalized?

skrthomas:
https://redhat-documentation.github.io/supplementary-style-guide/#ingress

mportman12:
@skrthomas Based on the context, I think it would be useful to add a sentence to the Ingress entry to note the non-capitalized usage. Could you open a PR for this?

skrthomas:
@mportman12 Yep, I can do that :) will post my PR here in the comments.

bergerhoffer:
@skrthomas I remember there being something different about "ingress", but I'm still not sure if I understand why "Ingress" is capitalized sometimes. I think it makes sense when it's the "Ingress Controller", but not just by itself. (I know it's listed capitali in the OCP glossary too, but that seems wrong to me

If it's the Ingress resource, then I would think it should be marked up as the `Ingress` resource, just like all of our other API objects on OpenShift. And then be lowercase for any general references.

In that case, shouldn't the entry just be lowercase "ingress" then? Like "config map" [1] and "pod" [2]?

[1] https://redhat-documentation.github.io/supplementary-style-guide/#config-map
[2] https://redhat-documentation.github.io/supplementary-style-guide/#pod

skrthomas:
@bergerhoffer I've always thought that when you're referring to an Ingress API resource, like the Ingress Controller, you capitalize. But if you're describing the direction of traffic, then ingress is lowercase. 

But yea, I agree, since my entry is not talking about the resource, I should lowercase it like the other entries you referenced.

skrthomas:
@bergerhoffer sorry, I realize that my entry is lowercase. 

My whole reason for submitting this is because from time to time, I have to talk about ingress traffic in the Network Observability documentation where there's a UI and you have visibility of ingress and egress traffic. Sometimes I get peer review questions wondering if I should capitalize it, pointing to the current SSG Ingress definition, which only defines the noun. So I was thinking that I could clarify this confusion by adding an entry for the ingress adjective, which describes the directional flow of network traffic vs the Ingress resource which can be used to control that flow.

bergerhoffer:
@skrthomas I think what's confusing is the fact that the ingress entry in the SSG is capitalized, and that's what needs to just be changed to lowercase.

[...]

PR #368, opened 2023-07-26
Adding ingress(adj) to the SSG

skrthomas:
https://github.com/redhat-documentation/supplementary-style-guide/issues/366

### DIFF
--- a/supplementary_style_guide/glossary_terms_conventions/general_conventions/i.adoc
+++ b/supplementary_style_guide/glossary_terms_conventions/general_conventions/i.adoc
@@ -323,12 +323,13 @@ There is no functional difference between the first server that was installed an
 *See also*:
 
 [[ingress]]
-==== image:images/yes.png[yes] Ingress (noun)
-*Description*: In Red Hat OpenShift, _Ingress_  is an API object developers can use to expose services through an HTTP(S) aware load balancing and proxy layer through a public DNS entry. The Ingress resource might further specify TLS options and a certificate, or specify a public CNAME that the OpenShift Ingress Controller should also accept for HTTP and HTTPS traffic. An administrator typically configures their Ingress Controller to be visible outside the cluster firewall and might also add additional security, caching, or traffic controls on the service content.
+==== image:images/yes.png[yes] ingress (noun, adjective)
+*Description*: In Red Hat OpenShift, _Ingress_  is a Kubernetes API object that developers can use to expose services within the cluster through an HTTP(S) load balancer and a proxy layer by using a public DNS entry. The `Ingress` resource defines the cluster-wide configuration for ingress traffic, and provides the ability to specify TLS options, a certificate, or a public CNAME that the OpenShift `IngressController` object can accept for HTTP(S) traffic. Additionally, _ingress_ can also be used to describe the incoming direction of network traffic. In Red Hat OpenShift, for example, this traffic is described as entering (ingress) or leaving (egress) an OpenShift cluster.
 
 *Use it*: yes
 
 *Incorrect forms*:
+_Ingress_ without markup, when referencing the `Ingress` resource or the `IngressController` object. _Ingress_ with markup when used to describe the direction of traffic.
 
 *See also*:
 


skrthomas:
@mportman12 here's my PR for #366

supplementary_style_guide/glossary_terms_conventions/general_conventions/i.adoc
```diff
@@ -324,12 +324,22 @@ There is no functional difference between the first server that was installed an
...
 ==== image:images/yes.png[yes] Ingress (noun)
-*Description*: In Red Hat OpenShift, _Ingress_  is an API object developers can use to expose services through an HTTP(S) aware load balancing and proxy layer through a public DNS entry. The Ingress resource might further specify TLS options and a certificate, or specify a public CNAME that the OpenShift Ingress Controller should also accept for HTTP and HTTPS traffic. An administrator typically configures their Ingress Controller to be visible outside the cluster firewall and might also add additional security, caching, or traffic controls on the service content.
+*Description*: In Red Hat OpenShift, _Ingress_  is an API object developers can use to expose services through an HTTP(S) aware load balancing and proxy layer through a public DNS entry. The `Ingress` resource might further specify TLS options and a certificate, or specify a public CNAME that the OpenShift Ingress Controller should also accept for HTTP and HTTPS traffic. An administrator typically configures their Ingress Controller to be visible outside the cluster firewall and might also add additional security, caching, or traffic controls on the service content.
```
skrthomas, line 327: @bergerhoffer I added markup for mentionings of the `Ingress` resource. I also added some context to the *Incorrect forms* description. Wdyt? Does this clarify the ambiguity a little further? Any other things that are a little unclear still? 🤔

supplementary_style_guide/glossary_terms_conventions/general_conventions/i.adoc
```diff
@@ -324,12 +324,22 @@ There is no functional difference between the first server that was installed an
...
 
 *Use it*: yes
 
 *Incorrect forms*:
+"Ingress" alone, when referencing the `Ingress` resource.
+*See also*:
+
+[[ingress]]
+==== image:images/yes.png[yes] ingress (adjective)
+*Description*: In addition to the API object, ingress also describes the incoming direction of network traffic. In Red Hat OpenShift, this traffic is often entering (ingress) or leaving (egress) an OCP cluster. The Ingress resource described above defines rules for how ingress network traffic should be routed from external sources into your services. An Ingress Controller is responsible for implementing these rules.
```
mportman12, line 337:
```suggestion
*Description*: _Ingress_ describes the incoming direction of network traffic. In Red Hat OpenShift, this traffic is often entering (ingress) or leaving (egress) an OCP cluster. The Ingress resource defines rules for how ingress network traffic should be routed from external sources into your services. An Ingress Controller is responsible for implementing these rules.
```

supplementary_style_guide/glossary_terms_conventions/general_conventions/i.adoc
```diff
@@ -324,12 +324,22 @@ There is no functional difference between the first server that was installed an
...
+*Description*: In Red Hat OpenShift, _Ingress_  is an API object developers can use to expose services through an HTTP(S) aware load balancing and proxy layer through a public DNS entry. The `Ingress` resource might further specify TLS options and a certificate, or specify a public CNAME that the OpenShift Ingress Controller (`IngressController`) should also accept for HTTP and HTTPS traffic. An administrator typically configures their Ingress Controller to be visible outside the cluster firewall and might also add additional security, caching, or traffic controls on the service content.
 
 *Use it*: yes
 
 *Incorrect forms*:
+"Ingress" alone, when referencing the `Ingress` resource.
```
mportman12, line 332: This is confusing because the entry is for "Ingress". Move to the description.

supplementary_style_guide/glossary_terms_conventions/general_conventions/i.adoc
```diff
@@ -324,12 +324,22 @@ There is no functional difference between the first server that was installed an
...
-*Description*: In Red Hat OpenShift, _Ingress_  is an API object developers can use to expose services through an HTTP(S) aware load balancing and proxy layer through a public DNS entry. The Ingress resource might further specify TLS options and a certificate, or specify a public CNAME that the OpenShift Ingress Controller should also accept for HTTP and HTTPS traffic. An administrator typically configures their Ingress Controller to be visible outside the cluster firewall and might also add additional security, caching, or traffic controls on the service content.
+*Description*: In Red Hat OpenShift, _Ingress_  is an API object developers can use to expose services through an HTTP(S) aware load balancing and proxy layer through a public DNS entry. The `Ingress` resource might further specify TLS options and a certificate, or specify a public CNAME that the OpenShift Ingress Controller (`IngressController`) should also accept for HTTP and HTTPS traffic. An administrator typically configures their Ingress Controller to be visible outside the cluster firewall and might also add additional security, caching, or traffic controls on the service content.
```
mportman12, line 327: From the context, it's not clear whether the Ingress API object and the Ingress resource are two different entities or the same entity.

supplementary_style_guide/glossary_terms_conventions/general_conventions/i.adoc
```diff
@@ -324,12 +324,22 @@ There is no functional difference between the first server that was installed an
...
 
 *Use it*: yes
 
 *Incorrect forms*:
+"Ingress" alone, when referencing the `Ingress` resource.
+*See also*:
+
+[[ingress]]
```
mportman12, line 335:
```suggestion
[[ingress-adj]]
```
Each glossary entry needs to have a unique ID to ensure that anchor links work correctly.

supplementary_style_guide/glossary_terms_conventions/general_conventions/i.adoc
```diff
@@ -324,12 +324,22 @@ There is no functional difference between the first server that was installed an
...
+*Description*: In Red Hat OpenShift, _Ingress_  is an API object developers can use to expose services through an HTTP(S) aware load balancing and proxy layer through a public DNS entry. The `Ingress` resource might further specify TLS options and a certificate, or specify a public CNAME that the OpenShift Ingress Controller (`IngressController`) should also accept for HTTP and HTTPS traffic. An administrator typically configures their Ingress Controller to be visible outside the cluster firewall and might also add additional security, caching, or traffic controls on the service content.
 
 *Use it*: yes
 
 *Incorrect forms*:
+"Ingress" alone, when referencing the `Ingress` resource.
+*See also*:
```
mportman12, line 333:
```suggestion
*See also*: xref:ingress-adj[ingress]
```

supplementary_style_guide/glossary_terms_conventions/general_conventions/i.adoc
```diff
@@ -324,12 +324,22 @@ There is no functional difference between the first server that was installed an
...
+
+[[ingress]]
+==== image:images/yes.png[yes] ingress (adjective)
+*Description*: In addition to the API object, ingress also describes the incoming direction of network traffic. In Red Hat OpenShift, this traffic is often entering (ingress) or leaving (egress) an OCP cluster. The Ingress resource described above defines rules for how ingress network traffic should be routed from external sources into your services. An Ingress Controller is responsible for implementing these rules.
 
+*Use it*: yes
+
+*Incorrect forms*:
+Ingress used to describe network traffic direction does not need to be capitalized or marked up like an `Ingress` resource.
 *See also*:
```
mportman12, line 343:
```suggestion
*See also*: xref:ingress[Ingress]
```

mportman12 (changes requested):
Left some suggestions and comments.

mportman12:
@skrthomas Please review suggested changes when you have a chance.

supplementary_style_guide/glossary_terms_conventions/general_conventions/i.adoc
bergerhoffer: @skrthomas Sorry for the late reply. The markup around `Ingress` makes sense to me. But the entry for capitalized "Ingress" still doesn't.

Per pretty much every other OCP object we have, we have the actual API object marked up when referencedd (`Pod`), but then when we speak of it generically, it's lowercase (pod). So this uppercase "Ingress" entry doesn't make sense to me. It doesn't seem like there should be a reason that "Ingress" is used but not marked up.

Or... is that the problem? Is there really no legitimate use of when "ingress"/"Ingress" should be used as a generic reference? Should it either be the `Ingress` API object (marked up), or otherwise it needs to be "ingress" as an adjective, in front of a noun?

skrthomas: @bergerhoffer 
> It doesn't seem like there should be a reason that "Ingress" is used but not marked up.

The reason ingress could be used but not marked up would be when it is used to describe the direction of traffic flow, which is why I was thinking it could be its own entry in the Glossary. But I guess, further thinking on this, I'm wondering if we can keep a single "ingress" definition, and clarify that the Ingress resource exposes HTTP and HTTPS routes from outside the cluster to _services within the cluster_, basically routing incoming traffic to different cluster applications. 

Then we can make it clearer that you can refer to the `Ingress` object, which must be marked up, or you can describe the general directional flow of incoming traffic, which does not need to be marked up and can be lowercase. 

> Is there really no legitimate use of when "ingress"/"Ingress" should be used as a generic reference? Should it either be the Ingress API object (marked up), or otherwise it needs to be "ingress" as an adjective, in front of a noun?

In my experience writing about Networking, I've only ever used it to either reference the `Ingress Controller` resource or to describe directional flow of traffic. Maybe these are the two use cases similar to `Pod` and pod.

supplementary_style_guide/glossary_terms_conventions/general_conventions/i.adoc
skrthomas: I changed this to Ingress without markup when referencing the `Ingress` resource or `IngressController`

supplementary_style_guide/glossary_terms_conventions/general_conventions/i.adoc
skrthomas: @mportman12 Ingress API object and Ingress resource are the same thing, but I don't think interchanging them in this definition does any good. Probably best to just stick with Ingress object. I also think more can be done in this definition to draw the connection between the API object, the Ingress Controller, and the Ingress Operator while we're at giving this a rework. I used the context here to help clarify this definition: https://docs.openshift.com/container-platform/4.12/networking/ingress-operator.html#nw-installation-ingress-config-asset_configuring-ingress

supplementary_style_guide/glossary_terms_conventions/general_conventions/i.adoc
skrthomas: I was thinking actually maybe the meaning I'm trying to convey can actually be wrapped into one ingress entry, not creating a separate entry for the adjective use.

skrthomas:
@bergerhoffer @mportman12 Apologies, I had a GA last Thursday and I've been bogged by that for a good month and a half and haven't had a chance to get back to this until now. I made some adjustments to my PR based on your feedback and questions (THANK YOU!!). I was thinking, maybe the ingress (noun) definition just needs a little more clarification and added detail to encapsulate the adjective version of the word I was trying to define here. I added more clarifying details to the existing definition in lieu of a new ingress adjective entry.

supplementary_style_guide/glossary_terms_conventions/general_conventions/i.adoc
```diff
@@ -323,13 +323,13 @@ There is no functional difference between the first server that was installed an
...
+==== image:images/yes.png[yes] ingress (noun) (adjective)
+*Description*: In Red Hat OpenShift, _Ingress_  is an API object that developers can use to expose services within the cluster through an HTTP(S) aware load balancer and a proxy layer via a public DNS entry. The `Ingress` resource defines the cluster-wide configuration for ingress traffic, and provides the ability to specify TLS options, a certificate, or a public CNAME that the OpenShift Ingress Controller (`IngressController`) can accept for HTTP(S) traffic. The OpenShift Ingress Operator uses the domain from the cluster `Ingress` object configuration as the domain for the default `IngressController`. An administrator typically configures the  `IngressController` to be visible outside the cluster firewall and might also add additional security, caching, or traffic controls on the service content. Furthermore, ingress also describes the incoming direction of network traffic. In Red Hat OpenShift, this traffic is described as entering (ingress) or leaving (egress) an OCP cluster.
```
mportman12, line 327:
```suggestion
*Description*: In Red Hat OpenShift, _Ingress_  is an API object that developers can use to expose services within the cluster through an HTTP(S) aware load balancer and a proxy layer using a public DNS entry. The `Ingress` resource defines the cluster-wide configuration for ingress traffic, and provides the ability to specify TLS options, a certificate, or a public CNAME that the OpenShift Ingress Controller (`IngressController`) can accept for HTTP(S) traffic. Furthermore, _ingress_ also describes the incoming direction of network traffic. In Red Hat OpenShift, this traffic is described as entering (ingress) or leaving (egress) an OCP cluster.
```
The description is very long and detailed - not sure it needs to be so detailed.

mportman12 (changes requested):

skrthomas:
@mportman12 @bergerhoffer how do these updates look now?

mportman12 (dismissed):
LGTM

supplementary_style_guide/glossary_terms_conventions/general_conventions/i.adoc
```diff
@@ -323,13 +323,13 @@ There is no functional difference between the first server that was installed an
 *See also*:
 
 [[ingress]]
-==== image:images/yes.png[yes] Ingress (noun)
-*Description*: In Red Hat OpenShift, _Ingress_  is an API object developers can use to expose services through an HTTP(S) aware load balancing and proxy layer through a public DNS entry. The Ingress resource might further specify TLS options and a certificate, or specify a public CNAME that the OpenShift Ingress Controller should also accept for HTTP and HTTPS traffic. An administrator typically configures their Ingress Controller to be visible outside the cluster firewall and might also add additional security, caching, or traffic controls on the service content.
+==== image:images/yes.png[yes] ingress (noun) (adjective)
```
bredamc, line 326:
```suggestion
==== image:images/yes.png[yes] ingress (noun, adjective)
```

```suggestion
==== image:images/yes.png[yes] ingress
```

It seems odd to me to have two sets of parentheses. I suggest either putting both parts of speech inside a single set of parentheses, or omit the parts of speech altogether. Or split them into two entries, one for noun and one for adjective.

supplementary_style_guide/glossary_terms_conventions/general_conventions/i.adoc
```diff
@@ -323,13 +323,13 @@ There is no functional difference between the first server that was installed an
...
+*Description*: In Red Hat OpenShift, _Ingress_  is an API object that developers can use to expose services within the cluster through an HTTP(S) aware load balancer and a proxy layer using a public DNS entry. The `Ingress` resource defines the cluster-wide configuration for ingress traffic, and provides the ability to specify TLS options, a certificate, or a public CNAME that the OpenShift Ingress Controller (`IngressController`) can accept for HTTP(S) traffic. Additionally, _ingress_ can also be used to describe the incoming direction of network traffic. In Red Hat OpenShift, for example, this traffic is described as entering (ingress) or leaving (egress) an OCP cluster.
```
bredamc, line 327:
```suggestion
*Description*: In Red Hat OpenShift, _Ingress_  is an API object that developers can use to expose services within the cluster through an HTTP(S) aware load balancer and a proxy layer using a public DNS entry. The `Ingress` resource defines the cluster-wide configuration for ingress traffic, and provides the ability to specify TLS options, a certificate, or a public CNAME that the OpenShift Ingress Controller (`IngressController`) can accept for HTTP(S) traffic. Additionally, _ingress_ can also be used to describe the incoming direction of network traffic. In Red Hat OpenShift, for example, this traffic is described as entering (ingress) or leaving (egress) an OCP cluster.
```

1. Suggest putting each sentence on a new line to make it easier for reviewers to add comments :)
2. Is the first sentence necessary? We should avoid using "Ingress" on its own. 
3. What noun should be used after `IngressController`? Is it a resource or an object or something else?
4. "OCP" is not an approved abbreviation, do not use.
5. Should we add an entry for "egress", for completeness?

bredamc (commented):
@skrthomas A few suggestions and questions for you to consider :)

supplementary_style_guide/glossary_terms_conventions/general_conventions/i.adoc
skrthomas: Thanks @bredamc , sorry for being so delayed getting back to this! Initially, I created a separate adjective entry, but initial reviews found that more confusing. So I added the adjective ingress definition to the noun ingress definition. I think I'll just leave it as is, and add the comma between the parts of speech.

supplementary_style_guide/glossary_terms_conventions/general_conventions/i.adoc
skrthomas: 1. ACK, good to know. Since this has already gone through most of the reviews its going to have, I think I'll forego and just make sure to do this for future style guide contributions :) Thanks for the tip though!
2. The first sentence is necessary, and I think it should be clarified to say that "_ingress_ is a Kubernetes API object". I was just following the convention of italicizing the term being defined, although you are correct that technically this usage of Ingress object would typically get backticks. 
3. resource. added!
4. ACK. Changed to OpenShift. 
5. That's a good idea. I would prefer to handle that addition in a separate PR, just so I can go ahead and merge this one. I opened it in July 😓

skrthomas:
@bredamc thanks for your review, and apologies for the delay here. I had some other feature work and bugs take my attention but I think I've addressed your feedback. I can handle the egress addition in a separate PR entry.

@bergerhoffer @mportman12 I think this needs re-reviewed by you since I pushed my updates for Breda's feedback. TY!! I hate that this has been open since July 😓 ...

mportman12 (approved):
LGTM

supplementary_style_guide/glossary_terms_conventions/general_conventions/i.adoc
```diff
@@ -323,13 +323,13 @@ There is no functional difference between the first server that was installed an
...
+==== image:images/yes.png[yes] ingress (noun, adjective)
+*Description*: In Red Hat OpenShift, _Ingress_  is a Kubernetes API object that developers can use to expose services within the cluster through an HTTP(S) aware load balancer and a proxy layer using a public DNS entry. The `Ingress` resource defines the cluster-wide configuration for ingress traffic, and provides the ability to specify TLS options, a certificate, or a public CNAME that the OpenShift `IngressController` resource can accept for HTTP(S) traffic. Additionally, _ingress_ can also be used to describe the incoming direction of network traffic. In Red Hat OpenShift, for example, this traffic is described as entering (ingress) or leaving (egress) an OpenShift cluster.
```
dfitzmau, line 327: Thanks, @skrthomas , for adding this item. I've queried these trms in the past as they are highlighted illogically sometimes in customer documentation. 

Just wondering about "HTTP(S) aware load balance": I understand with the paranethesis, appending "aware" with a hyphen, which would be common, might be difficult, so how about: "an HTTP(S) load balancer" to drop the potential hyphenated construct?

Expand CNAME?

"IngressController object" more precise? I see the object reference commonly adopted in OCP docs. 

The "See also" subheading needs to be escaped onto the next line.

supplementary_style_guide/glossary_terms_conventions/general_conventions/i.adoc
```diff
@@ -323,13 +323,13 @@ There is no functional difference between the first server that was installed an
...
 
 *Use it*: yes
 
 *Incorrect forms*:
-
+Ingress without markup, when referencing the `Ingress` resource or `IngressController`.
```
dfitzmau, line 332:
```suggestion
"Ingress" without markup When referencing the `Ingress` resource or `IngressController`.
```

supplementary_style_guide/glossary_terms_conventions/general_conventions/i.adoc
```diff
@@ -323,13 +323,13 @@ There is no functional difference between the first server that was installed an
...
 *See also*:
```
dfitzmau, line 333: Try adding a space before "See also" to avoid it falling back on the previous line.

supplementary_style_guide/glossary_terms_conventions/general_conventions/i.adoc
IngridT1: @dfitzmau I had the same question about whether or not to hyphenate "HTTP(S) aware load balance". If removing "aware" works, that would solve the issue. @skrthomas wdyt?

IngridT1 (commented):
LGTM if the comments are addressed.

supplementary_style_guide/glossary_terms_conventions/general_conventions/i.adoc
skrthomas: I think dropping the "aware" is totally fine. Good catch :)  It was a holdover from the original definition that I didn't change but could definitely be changed without altering the meaning.

skrthomas: >Expand CNAME?
>"IngressController object" more precise? I see the object reference commonly adopted in OCP docs.

I do not think CNAME needs expanding.
Yea, I think `Ingress` resource and `IngressController` object is precise. thanks for pointing this out!

skrthomas:
@IngridT1 @dfitzmau I've incorporated your feedback. Thanks for your reviews! I guess it needs another approval before its good to go? 

cc @mportman12

supplementary_style_guide/glossary_terms_conventions/general_conventions/i.adoc
```diff
@@ -323,12 +323,13 @@ There is no functional difference between the first server that was installed an
...
+==== image:images/yes.png[yes] ingress (noun, adjective)
+*Description*: In Red Hat OpenShift, _Ingress_  is a Kubernetes API object that developers can use to expose services within the cluster through an HTTP(S) load balancer and a proxy layer using a public DNS entry. The `Ingress` resource defines the cluster-wide configuration for ingress traffic, and provides the ability to specify TLS options, a certificate, or a public CNAME that the OpenShift `IngressController` object can accept for HTTP(S) traffic. Additionally, _ingress_ can also be used to describe the incoming direction of network traffic. In Red Hat OpenShift, for example, this traffic is described as entering (ingress) or leaving (egress) an OpenShift cluster.
```
IngridT1, line 327: Sorry, I just notice a small issue. Could you please add a word to this sentence?

> In Red Hat OpenShift, _Ingress_ is a Kubernetes API object that developers can use to expose services within the cluster through an HTTP(S) load balancer and a proxy layer using a public DNS entry.

Please add "by" before "using a public DNS entry."

joaedwar (dismissed):
lgtm

supplementary_style_guide/glossary_terms_conventions/general_conventions/i.adoc
skrthomas: Addressed :)

dfitzmau: The timing is excellent as I just stumbled across CNAME. It means "Canonical Name". Maybe put this in parentheses? JUst more a case of keeping consistent with spelling out full terms on first use.

skrthomas: When I search the OCP docs repo, CNAME is never expanded in parentheses so I don't really know if its necessary to do that here in this definition.

dfitzmau: @redhat-documentation/ccs-style-council , WDYT? OUt of context or good for consistency of rules?

mportman12: I vote out of context

dfitzmau (approved):

mportman12 (approved):
LGTM

Issue #147, opened 2022-01-05
Fix various style issues and typos throughout the guide 


bredamc:
- [x] 1. General: Several of the words listed are already included in IBM Style. Examples include "agnostic", "client side", "cloud", "colocate", "data center", "DevOps", and so on. It might be a good idea to compare the list with the latest version of the "Word usage" topic (https://www.ibm.com/docs/en/ibm-style?topic=word-usage), and update the IBM Style entries as necessary? **[Volunteer needed]**

- [x] 2. General: Search for occurrences of "UNIX" to ensure that it is always written in all uppercase letters (see the entry for "UNIX" in the "Word usage" topic https://www.ibm.com/docs/en/ibm-style?topic=word-usage#u). **[Breda McColgan]**
https://github.com/redhat-documentation/supplementary-style-guide/pull/163

- [x] 3. General: Search for occurrences of "may" that should be changed to "can" or "might" -- see the entry for "may" in the "Word usage" topic (https://www.ibm.com/docs/en/ibm-style?topic=word-usage#m). **[Breda McColgan]**
https://github.com/redhat-documentation/supplementary-style-guide/pull/166

- [x] 4. General: Search for occurrences of "acronym" to ensure that they are all actual acronyms according to the IBM Style definition (https://www.ibm.com/docs/en/ibm-style?topic=grammar-abbreviations) -- many are initialisms, which are not pronounced as words. **[Breda McColgan]**
https://github.com/redhat-documentation/supplementary-style-guide/pull/166

- [ ] 5. General: Be consistent in the descriptions of related words:
    (a) Example 1: KB, MB, GB (see https://www.ibm.com/docs/en/ibm-style?topic=measurement-units#si-multiplier-prefixes and https://www.ibm.com/docs/en/ibm-style?topic=measurement-units#multiplier-prefixes-for-bits-and-bytes).
    (b) Example 2: online backup, offline backup.
    In any case, these entries might not be needed -- IBM Style already has entries for "online" and "offline" (see https://www.ibm.com/docs/en/ibm-style?topic=word-usage#o). 
  **[Volunteer needed]**

- [x] 6. https://redhat-documentation.github.io/supplementary-style-guide/#lead-in-sentences
Typo: "prerequisties". **[Breda McColgan]**
https://github.com/redhat-documentation/supplementary-style-guide/pull/166

- [ ] 7.  https://redhat-documentation.github.io/supplementary-style-guide/#release-notes
[...]

PR #391, opened 2023-09-07
fix: revise last sentence in paragraph about segmentation-fault

CBID2:
## Description
This PR updates the description on how to use a segmentation fault, ensuring that the description is clear and follows the established style guide.

## Issue
Closes number 21 in #147


.DS_Store
bergerhoffer, line 1: This extraneous file needs to be removed (there is also another one inside of supplementary_style_guide)

bergerhoffer (commented):
There are two `.DS_Store` files that were included here that need to be removed. But once those are removed, the actual update LGTM!

CBID2:
> There are two `.DS_Store` files that were included here that need to be removed. But once those are removed, the actual update LGTM!

Hi @bergerhoffer. The files are gone now! :)

bergerhoffer (approved):
LGTM!

Just needs 2 more approvals @redhat-documentation/ccs-style-council

bredamc:
@CBID2 Many thanks for making so many of the updates that I suggested in that issue, much appreciated!

I know you have made the change that I originally suggested (thank you!), but now I think we should also append the sentence "Never use the abbreviation "segfault" as a verb." -- I think that new sentence makes that point more strongly than the text I'd originally suggested.

CBID2:
> @CBID2 Many thanks for making so many of the updates that I suggested in that issue, much appreciated!
> 
> 
> 
> I know you have made the change that I originally suggested (thank you!), but now I think we should also append the sentence "Never use the abbreviation "segfault" as a verb." -- I think that new sentence makes that point more strongly than the text I'd originally suggested.

You're welcome @bredamc. I think I got rid of that part when I made the PR.

supplementary_style_guide/glossary_terms_conventions/general_conventions/s.adoc
```diff
@@ -164,7 +164,7 @@ Specify the system architecture of your cluster, such as `s390x` or `x86_64`.
 
 [[segmentation-fault]]
 ==== image:images/yes.png[yes] segmentation fault (noun)
-*Description*:  A _segmentation fault_ occurs when a process tries to access a memory location that it is not allowed to access, or tries to access a memory location in a way that is not allowed (for example, if the process tries to write to a read-only location or to overwrite part of the operating system). Only use the abbreviation "segfault" if absolutely necessary, and never use it as a verb.
+*Description*:  A _segmentation fault_ occurs when a process tries to access a memory location that it is not allowed to access, or tries to access a memory location in a way that is not allowed (for example, if the process tries to write to a read-only location or to overwrite part of the operating system). Do not use the abbreviation "segfault" unless absolutely necessary.
```
bredamc, line 167:
```suggestion
*Description*:  A _segmentation fault_ occurs when a process tries to access a memory location that it is not allowed to access, or tries to access a memory location in a way that is not allowed (for example, if the process tries to write to a read-only location or to overwrite part of the operating system). Do not use the abbreviation "segfault" unless absolutely necessary. Never use the abbreviation "segfault" as a verb.
```

bredamc: I had originally suggested removing that text (about the verb) but I think we should retain it.

bredamc (approved):
LGTM! Many thanks for the prompt updates :)

CBID2:
> @CBID2 Many thanks for making so many of the updates that I suggested in that issue, much appreciated!
> 
> I know you have made the change that I originally suggested (thank you!), but now I think we should also append the sentence "Never use the abbreviation "segfault" as a verb." -- I think that new sentence makes that point more strongly than the text I'd originally suggested.

@bredamc, I think this part "when a process tries to access a memory location that it is not allowed to access, or tries to access a memory location in a way that is not allowed" could be improved. It sounds a bit repetitive.
I was thinking of something like this:
A _segmentation fault_ occurs when a process tries to access a memory location that it is not allowed to access or uses an unauthorized method to gain information from said area, (for example, if the process tries to write to a read-only location or to overwrite part of the operating system). Do not use the abbreviation "segfault" unless absolutely necessary. Never use the abbreviation "segfault" as a verb.

bredamc:
@CBID2 Many thanks for your suggested update. I actually prefer the original wording because it clarifies that either the _location_ or the _access method_ might be the problem. But I'm not an expert in this area so I'll defer to @bergerhoffer -- she might know who wrote the original text. :)

CBID2:
> @CBID2 Many thanks for your suggested update. I actually prefer the original wording because it clarifies that either the _location_ or the _access method_ might be the problem. But I'm not an expert in this area so I'll defer to @bergerhoffer -- she might know who wrote the original text. :)

Oh in that case @bredamc, that's fine. Let's merge this for now.

CBID2:
Hey @bergerhoffer. I made the changes

bergerhoffer (approved):

mportman12 (approved):
LGTM

Issue #393, opened 2023-09-07
Add "Application" to Homographs


jafiala:
I propose adding "Application" to the list of homographs, because this might actually be a rather common occurrence.

For example, just came across this sentence in a draft during peer review: "...choosing between application stability and application of security updates."

Also, while at it, let's capitalize all items in the Homographs list to be compliant with [IBM Style (Capitalize the first word of each item in a vertical list.) ](https://www.ibm.com/docs/en/ibm-style?topic=format-lists#capitalization-in-lists)

bergerhoffer:
No objections from me to add "Application" and to capitalize this list items.

@jafiala would you be able to open a PR for this?

jafiala:
Done, please review: https://github.com/redhat-documentation/supplementary-style-guide/pull/394

bergerhoffer:
Closing because PR #394 has been merged

PR #394, opened 2023-09-08
Add application to homographs, harmonize capitalization in lists

jafiala:
Add "application" to the list of homographs
Consistently apply initial capitalization to lists in the grammar chapter to be stylistically consistent

https://github.com/redhat-documentation/supplementary-style-guide/issues/393#issuecomment-1710359297

### DIFF
--- a/supplementary_style_guide/style_guidelines/grammar.adoc
+++ b/supplementary_style_guide/style_guidelines/grammar.adoc
@@ -12,14 +12,14 @@ For more information about the Conscious Language Group, see https://github.com/
 To ensure consistency and success, it is imperative for product team stakeholders to align internally. For example, documentation teams should engage in discussions with their engineering leadership to reach an agreement on replacement terms. This ensures that the product documentation matches the code.
 ====
 
-=== blacklist and whitelist
+=== Blacklist and whitelist
 
 When possible, rewrite documentation to avoid these terms.
 When it is not possible to remove the terms _blacklist_ and _whitelist_, replace them with one of the following alternatives:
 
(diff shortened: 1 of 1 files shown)


IngridT1:
I think that we should change the definition of homograph slightly. Right now, it says this:

> A homograph is a word that is spelled the same as another word but has a different 
> meaning when used as a different part of speech. 

"Application" is always a noun in all its different meanings. How about this definition instead?

> A homograph is a word that shares the same written form as another word but has a different meaning.

Wikipedia has a nuanced discussion of homographs, and "application" doesn't fit every definition of homograph: https://en.wikipedia.org/wiki/Homograph

But because "application" is a common word in our field, it makes sense to include it on the list.

jafiala:
Thanks Ingrid, that's a good point.

I agree that we should clarify, and I like your wording.

Not sure if I should add this modification to the same PR though.

sbmetz (dismissed):
LGTM

dfitzmau (dismissed):

bredamc:
I think if you're going to add "Application" to the list, you need to change the definition of "homograph" in that section. I think the simplest way to edit is to just remove the final 8 words, so that the definition is as follows:

> A homograph is a word that is spelled the same as another word but has a different meaning.

jafiala:
Hey @bredamc , removed these words (and also added "also" to make the sentence work).

bredamc:
Thanks @jafiala. I think it's better without the "also".

bredamc (approved):
LGTM! Thanks @jafiala!

bburt-rh:
lgtm

bburt-rh (approved):

bergerhoffer (approved):
LGTM!

bergerhoffer:
3 approvals - merging!

PR #452, opened 2024-03-01
Added a description for using the word currently

### LINKED ISSUE #428: Review and updated temporal language for known issues?
In https://github.com/redhat-documentation/supplementary-style-guide/pull/425, @bergerhoffer raised the following point:

"Added some feedback but I'd also like to ask more about the "now" guidance. And whether this also applies to "currently", which is similar to "now". The examples in the release notes section for known issues use "currently".

The "before this update" replacement for "now" works for fixed bugs, but not for known issues. Should we provide a suggested now/currently replacement for known issues? Something like "As of this release" or similar?"

lburcher7:
I added "currently" to the glossary of terms.

### DIFF
--- a/supplementary_style_guide/glossary_terms_conventions/general_conventions/c.adoc
+++ b/supplementary_style_guide/glossary_terms_conventions/general_conventions/c.adoc
@@ -981,6 +981,17 @@ With a _cross-forest trust_ between an Active Directory (AD) forest root domain 
 
 *See also*:
 
+[[currently]]
+==== image:images/yes.png[yes] currently (adverb)
+*Description*: Use _currently_ to refer to a situation that will change in the future. For example, you can use _currently_ in release notes when a known issue will likely be fixed in an upcoming release. 
+
+*Use it*: yes
+
+[.vale-ignore]
+*Incorrect forms*: 
+
+*See also*:
+
 // OCP: Added "In Red Hat OpenShift,"
 [[custom-resource]]
 ==== image:images/yes.png[yes] custom resource (noun)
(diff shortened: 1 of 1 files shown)


bburt-rh (dismissed):
lgtm

supplementary_style_guide/glossary_terms_conventions/general_conventions/c.adoc
```diff
@@ -981,6 +981,17 @@ With a _cross-forest trust_ between an Active Directory (AD) forest root domain
 
 *See also*:
 
+[[currently]]
+==== image:images/yes.png[yes] currently (adverb)
+*Description*: Use _currently_ to refer to a situation that will change in the future. For example, you can use _currently_ in release notes when a known issue will likely be fixed in an upcoming release.
```
bergerhoffer, line 986: * Since we're telling how to use it, then it should be quotes
* Also softened "will change" to "might change", since I think it can still be currently even if it doesn't end up changing.

```suggestion
*Description*: Use "currently" to refer to a situation that might change in the future. For example, you can use "currently" in release notes when a known issue will likely be fixed in an upcoming release. 
```

bergerhoffer (commented):
Thanks for adding! Added a few minor suggestions

lburcher7:
Thank you, @bburt-rh and @bergerhoffer!

bergerhoffer (approved):
LGTM!

bburt-rh (approved):
LGTM

rolfedh (approved):
LGTM

Issue #512, opened 2024-08-13
Update Tech Preview and Dev Preview templates to better emphasize not to use in production environments


bburt-rh:
The Automotive docs team proposes to make the following updates to the [Tech Preview and Dev Preview templates in the SSG](https://redhat-documentation.github.io/supplementary-style-guide/#support) to emphasize and reinforce that that software released as TP or DP is not intended for production environments. The current wording places this warning in the middle of the text block, and the proposed updates move the warning closer to the beginning for emphasis.

Note that the Automotive docs team management has already gotten approval from Red Hat Legal for these changes.

**Current AsciiDoc Dev Preview template:**

```
[IMPORTANT]
====
<software_name> is Developer Preview software only. Developer Preview software is not supported by Red{nbsp}Hat in any way and is not functionally complete or production-ready. Do not use Developer Preview software for production or business-critical workloads. Developer Preview software provides early access to upcoming product software in advance of its possible inclusion in a Red{nbsp}Hat product offering. Customers can use this software to test functionality and provide feedback during the development process. This software might not have any documentation, is subject to change or removal at any time, and has received limited testing. Red{nbsp}Hat might provide ways to submit feedback on Developer Preview software without an associated SLA.

For more information about the support scope of Red{nbsp}Hat Developer Preview software, see link:https://access.redhat.com/support/offerings/devpreview/[Developer Preview Support Scope].
====

```

**Proposed AsciiDoc Dev Preview template:**

```
[IMPORTANT]
====
{ProductName} is Developer Preview software only, and as such, it is not intended for production or business-critical workloads. Developer Preview software is not supported by Red{nbsp}Hat in any way and is not functionally complete. This software is subject to change or removal at any time, and has received limited testing only.
Developer Preview software provides early access to upcoming product software in advance of its possible inclusion in a Red{nbsp}Hat product offering. Customers can use this software to test functionality and provide feedback during the development process. Red{nbsp}Hat might provide ways to submit feedback on Developer Preview software without an associated SLA.
For more information about the support scope of Red{nbsp}Hat Developer Preview software, see link:https://access.redhat.com/support/offerings/devpreview/[Developer Preview Support Scope].
====
```

**Current AsciiDoc Tech Preview template:**

```
[IMPORTANT]
====
[...]

PR #513, opened 2024-08-13
Issue-512: update-wording-for-tp-dp-templates

bburt-rh:
Issue: #512 

Additional information:
Updates the tech preview and dev preview templates to emphasize more clearly that these software releases are not intended for production environments.

### DIFF
--- a/supplementary_style_guide/style_guidelines/support.adoc
+++ b/supplementary_style_guide/style_guidelines/support.adoc
@@ -33,8 +33,10 @@ Use the following template. Replace _<software_name>_ with the software name:
 ----
 [IMPORTANT]
 ====
-_<software_name>_ is Developer Preview software only. Developer Preview software is not supported by Red{nbsp}Hat in any way and is not functionally complete or production-ready. Do not use Developer Preview software for production or business-critical workloads. Developer Preview software provides early access to upcoming product software in advance of its possible inclusion in a Red{nbsp}Hat product offering. Customers can use this software to test functionality and provide feedback during the development process. This software might not have any documentation, is subject to change or removal at any time, and has received limited testing. Red{nbsp}Hat might provide ways to submit feedback on Developer Preview software without an associated SLA.
-
+_<software_name>_ is Developer Preview software only, and as such, it is not intended for production or business-critical workloads. Developer Preview software is not supported by Red{nbsp}Hat in any way and is not functionally complete.
+This software is subject to change or removal at any time, and has received limited testing only.
+Developer Preview software provides early access to upcoming product software in advance of its possible inclusion in a Red{nbsp}Hat product offering. Customers can use this software to test functionality and provide feedback during the development process.
+Red{nbsp}Hat might provide ways to submit feedback on Developer Preview software without an associated SLA.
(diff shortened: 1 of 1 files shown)


supplementary_style_guide/style_guidelines/support.adoc
```diff
@@ -33,8 +33,10 @@ Use the following template. Replace _<software_name>_ with the software name:
 ----
 [IMPORTANT]
 ====
-_<software_name>_ is Developer Preview software only. Developer Preview software is not supported by Red{nbsp}Hat in any way and is not functionally complete or production-ready. Do not use Developer Preview software for production or business-critical workloads. Developer Preview software provides early access to upcoming product software in advance of its possible inclusion in a Red{nbsp}Hat product offering. Customers can use this software to test functionality and provide feedback during the development process. This software might not have any documentation, is subject to change or removal at any time, and has received limited testing. Red{nbsp}Hat might provide ways to submit feedback on Developer Preview software without an associated SLA.
-
+_<software_name>_ is Developer Preview software only, and as such, it is not intended for production or business-critical workloads. Developer Preview software is not supported by Red{nbsp}Hat in any way and is not functionally complete.
```
bergerhoffer, line 36: Right now, this is rendering as each of these lines on separate lines (unlike normal paragraphs where this would have been smushed together). 

So if it's supposed to be a single block paragraph, then you likely need to put these all on the same line.

But also should consider whether there should be any blank lines between them. At the very least, there should likely be a blank line before "For more information about..." to match how it was in the previous snippet text. But I think it might be a good time to decide whether any additional breaks could help readability

supplementary_style_guide/style_guidelines/support.adoc
```diff
@@ -68,8 +70,10 @@ Use the following template text verbatim, where _<feature_name>_ is your feature
 ----
 [IMPORTANT]
 ====
-_<feature_name>_ is a Technology Preview feature only. Technology Preview features are not supported with Red{nbsp}Hat production service level agreements (SLAs) and might not be functionally complete. Red{nbsp}Hat does not recommend using them in production. These features provide early access to upcoming product features, enabling customers to test functionality and provide feedback during the development process.
-
+_<feature_name>_ is a Technology Preview feature only.
```
bergerhoffer, line 73: Same comment about these being on individual lines when the preview is built

supplementary_style_guide/style_guidelines/support.adoc
```diff
@@ -68,8 +70,10 @@ Use the following template text verbatim, where _<feature_name>_ is your feature
...
+Technology Preview software is not intended for production or business-critical workloads.
```
bergerhoffer, line 74: I don't think we previously referred to it as "Technology Preview _software_". Just "Technology Preview _features_". Is that a purposeful change? If not, could we change it to say "Technology Preview features..."?

bergerhoffer (commented):
A few comments/questions!

bergerhoffer:
Did not have time at the 28 August SSG meeting to discuss this, due to the SSG stop activities announcement

PR #517, opened 2024-09-30
Add tech example: YAML ellipses

apinnick:
Issue: None

Additional information: Not covered in ISG.

PR #519, opened 2024-10-07
Ellipses yaml code blocks

apinnick:
Issue: None

Additional information: Review comments in https://github.com/redhat-documentation/supplementary-style-guide/pull/517


dfitzmau (approved):

ndeevy (approved):
LGTM

bredamc:
@apinnick Just to explain my request to show both the code and the rendered output for the example -- it's not clear to me whether the workaround of prefixing the ellipsis with ` #`  will render the ellipsis correctly in the output, or comment it out so that it is no longer visible?

apinnick:
> @apinnick Just to explain my request to show both the code and the rendered output for the example -- it's not clear to me whether the workaround of prefixing the ellipsis with ` #` will render the ellipsis correctly in the output, or comment it out so that it is no longer visible?

This is not a workaround. It's standard YAML. 

The `#` will always render because it is inside a code block.

bredamc (approved):
LGTM

PR #525, opened 2025-05-13
AI glossary first submission

kelbrown20:
This PR adds some AI terms to the style guide glossary


supplementary_style_guide/glossary_terms_conventions/general_conventions/i.adoc
```diff
@@ -359,6 +359,50 @@ There is no functional difference between the first server that was installed an
...
+*Description*: Inference is when a model can process, deduct, and produce outputs based on input data. The process of using a trained AI model to generate predictions or conclusions based on the input data provided to the model.
+
+*Use it*: yes
+
+[.vale-ignore]
+*Incorrect forms*: 
+
+*See also*:
+
+[[`InferenceService`]]
```
bergerhoffer, line 373: This is messing up the formatting

image

Should just be

```suggestion
[[inferenceservice]]
```

supplementary_style_guide/glossary_terms_conventions/general_conventions/i.adoc
```diff
@@ -359,6 +359,50 @@ There is no functional difference between the first server that was installed an
...
+
+*Use it*: yes
+
+[.vale-ignore]
+*Incorrect forms*: 
+
+*See also*:
+
+[[Inference Server]]
+==== image:images/yes.png[yes] Inference Server (noun)
```
bergerhoffer, line 396: This should move above "inference serving" and "InferenceService" to be alphabetical

supplementary_style_guide/glossary_terms_conventions/general_conventions/i.adoc
```diff
@@ -359,6 +359,50 @@ There is no functional difference between the first server that was installed an
...
+
+*Use it*: yes
+
+[.vale-ignore]
+*Incorrect forms*: 
+
+*See also*:
+
+[[`InferenceService`]]
+==== image:images/yes.png[yes] `InferenceService` (noun)
```
bergerhoffer, line 374: We don't use markup in the headings. Just in the description below. Though to be honest, I don't think we've historically really put object names in as glossary entries. Is this always written this way, or would it sometimes be written generically as the "inference service"?

If yes, then I would put "inference service" as the glossary entry, and then say something in the description like, if you're talking about the actual object, then write it as `InferenceService`.

For a glossary entry in a similar situation, see "pod", "persistent volume claim", "service", etc.

```suggestion
==== image:images/yes.png[yes] InferenceService (noun)
```

supplementary_style_guide/glossary_terms_conventions/general_conventions/i.adoc
```diff
@@ -359,6 +359,50 @@ There is no functional difference between the first server that was installed an
...
+*Description*: In Red Hat OpenShift AI, this is the Custom Resource  Definition (CRD) used to create the `InferenceService` object. Written as the CRD name `InferenceService` with backticks. 
+
+*Use it*: yes
+
+[.vale-ignore]
+*Incorrect forms*: InferenceService, inference serving, inferenceservice
```
bergerhoffer, line 380: Was this incorrect form supposed to be "inference service"?

(Also going back to my question about whether "inference service" could ever been used generically. It's fine if it's how you want to do it, but I also express my concerns about references to a certain concept always having to be in pascal case/monospace, since that's not a great reader experience)

```suggestion
*Incorrect forms*: InferenceService, inference serving, inference service 
```

bergerhoffer (commented):
Here's a first pass at a few bits of feedback!

supplementary_style_guide/glossary_terms_conventions/general_conventions/i.adoc
kelbrown20: I don't think we want to suggest 'inference service' on its own. I only see it in the context of the object, but ill ask the RHOAI folks as well.

supplementary_style_guide/glossary_terms_conventions/general_conventions/i.adoc
```diff
@@ -359,6 +359,50 @@ There is no functional difference between the first server that was installed an
...
+*Use it*: yes
+
+[.vale-ignore]
+*Incorrect forms*: 
+
+*See also*:
+
+[[Inference Server]]
+==== image:images/yes.png[yes] Inference Server (noun)
+*Description*: In reference to product name, Red Hat AI Inference Server or RHAIIS. Capitalize the first letters.
```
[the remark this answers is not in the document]
kelbrown20: Lemme check on this one, Im updating the PR with the other fixes. Thank you!

supplementary_style_guide/glossary_terms_conventions/general_conventions/i.adoc
```diff
@@ -359,6 +370,28 @@ There is no functional difference between the first server that was installed an
...
+
+*Use it*: yes
+
+[.vale-ignore]
+*Incorrect forms*: InferenceService, inference serving
+
+*See also*:
+
+[[inference-serving]]
+==== image:images/yes.png[yes] inference serving (verb)
```
aireilly, line 385: So is "inferencing" accepted verb usage for serving models? For example, "inferencing the quantized granite model".

kelbrown20: So this is a great point, early on when researching this, I rarely ever saw usage of the the verb form "inferencing" , but Ive been seeing it more and more and It seems like its its becoming standard. So I'll update this to so that "inferencing" can be used. I think part of the confusing part was it seemed like PMs were using "inference serving" primarily as the single word for both inferencing and serving, which still works. So I'll keep this term, update the description to make that more clear, and make "inferencing" an option

sbmetz: Seems jargony to me, vs "performing/running inference" or "inference serving". (ChatGPT and Gemini support this analysis :smiley: )

aireilly: Both appear widely used in academic papers.

supplementary_style_guide/glossary_terms_conventions/general_conventions/i.adoc
```diff
@@ -348,6 +348,28 @@ There is no functional difference between the first server that was installed an
...
+*Description*: The act a model generating outputs from input data. For example, "Inference speeds increased on the new models"
+
+*Use it*: yes
+
+[.vale-ignore]
+*Incorrect forms*: 
+
+*See also*:
+
+[[inferencing]]
```
aireilly:
```suggestion
[[inference serving]]
```

aireilly (commented):
Added some suggestions

supplementary_style_guide/glossary_terms_conventions/general_conventions/i.adoc
aireilly: "Inference serving" seems to be the proper form.

IngridT1: They seem to both be correct terms, but they have different meanings.

supplementary_style_guide/glossary_terms_conventions/general_conventions/i.adoc
```diff
@@ -359,6 +381,29 @@ There is no functional difference between the first server that was installed an
...
+    
+*Use it*: yes
+
+[.vale-ignore]
+*Incorrect forms*: InferenceService, inference serving
+
+*See also*:
+
+[[inference-serving]]
+==== image:images/yes.png[yes] inference serving (verb)
```
kelbrown20: We have both defined here, does this look like how we wanted to define `Inference serving` vs `inferencing`

IngridT1:
@aireilly I don't agree that "inferencing" is the same as "inference serving". @kelbrown20? 

I am not an AI expert by any means, but the definitions seem distinct to me. I'm thinking that we'll have to meet and hammer out final definitions synchronously, because 56 comments on a PR without a resolution is too many. 

I'm also tagging @bburt-rh because I'd like his thoughts. 

When I ask Gemini "Is there a difference between **inferencing** and **inference serving**," I get this response (truncated and rearranged somewhat):

> **Inferencing** is the actual mathematical process of a trained model making a prediction. It is the act of feeding input data into a neural network and getting an output. (The "What")
> 
> **Inference Serving** is the software infrastructure and "plumbing" that makes that model accessible to the world. It’s the framework that allows an application to send data to the model and receive an answer back reliably and at scale. (The "How")
> 
> While the terms are often used interchangeably in casual conversation, they represent two different layers of the machine learning lifecycle.
> 
> Think of **inferencing** as the calculation and **serving** as the service delivering it.

Issue #532, opened 2025-08-01
Update Code Block formatting section to add - Separate a command and its related example output into individual code blocks.


Preeticp:
1. Update guidance for formatting code blocks in the Supplementary Style Guide: https://redhat-documentation.github.io/supplementary-style-guide/#commands-in-code-blocks to add that the code that the user needs to input and the example output need to be in separate code blocks. 
Mixing the command and the output in a single code block copies the entire content in the code block and does not filter away the output as mentioned [here](https://github.com/redhat-documentation/supplementary-style-guide/issues/390) (The copy-paste button feature on the Customer Portal seems to be able to filter away the output...)
We could use the guidance provided [here](https://github.com/openshift/openshift-docs/blob/main/contributing_to_docs/doc_guidelines.adoc#source-tags-for-terminal-commands-and-output): "Separate a command and its related example output into individual code blocks." 

2. The example provided in the guide would also need to be updated (Example AsciiDoc: A command and its output in a code block).

So the end result could look like: 
1. Use a single command per code block for each procedure step. Separate a command and its related example output into individual code blocks. This approach helps with readability and makes it possible for the copy button in code blocks to work correctly.

2.  So the example could look like: 

```
[subs="+quotes"]
----
# *virsh net-list --all*
----
```

```
.Example output
[subs="+quotes"]
----
Name      State    Autostart   Persistent
--------------------------------------------
default   active   yes         yes
----
```

3.  This renders as:

Verify that the libvirt default network is active and configured to start automatically:

`**# virsh net-list --all**`

Example output
```
----
Name      State    Autostart   Persistent
--------------------------------------------
default   active   yes         yes
----
```

Preeticp:
Fixed through PR 534. Closing this issue.

PR #534, opened 2025-08-07
single_cmd_per_codeblock

Preeticp:
This PR addresses issue #532 and adds a line on separating a command and its related example output into individual code blocks and update examples.


jherrman:
For transparency purposes, there is a discussion on whether this change is actually warranted. The basic arguments against separating the command and the output in different codeblocks are the following:

* Clutters the source
* Clutters the rendered output
* Does not hinder the "Copy command" function (in most cases, at least)
* Would require updating a significant portion (possibly the majority) of our documentation in a way that might be difficult to automate

Issue #535, opened 2025-08-15
Current guidance to format commands as bold text by using asterisks breaks commands that already contain asterisk wildcards


rolfedh:
## Problem Description

The current SSG topic, [Commands in code blocks](https://redhat-documentation.github.io/supplementary-style-guide/#commands-in-code-blocks), prescribes using asterisks to format commands as bold text. 

Enclosing commands that already contain an asterisk in a pair of asterisks creates parsing conflicts that "break" the command. For example:

```asciidoc
$ *for I in $(oc get ns -o jsonpath='{range .items[*]} {.metadata.name}{"\n"} {end}'); do oc delete pods --all -n $I; done*
```

The first asterisk in the command (`items[*]`) terminates the bold formatting prematurely, causing the rest of the command to render incorrectly and exposing the trailing asterisk.

Screenshot of source:
<img width="765" height="175" alt="Image" src="https://github.com/user-attachments/assets/04d18604-1f8b-4ebe-9a87-2026ff4f4c91" />

Screenshot of output showing how the asterisk in `.items[*]` terminates the bold formatting, makes that asterisk disappear, and  exposes the asterisk at the end of the command.
<img width="702" height="139" alt="Image" src="https://github.com/user-attachments/assets/90a4ba0a-07ec-4b8f-b558-048e3edab83b" />

### 1. Escaping asterisks
```asciidoc
*$ for I in $(oc get ns -o jsonpath='{range .items[\\*]} ...')*
```
**Issues:**
- Requires manual escaping of every asterisk in the command
- Error-prone and difficult to maintain
- Reduces readability of source documentation

### 2. Attribute substitution
```asciidoc
:asterisk: *
$ *for I in $(oc get ns -o jsonpath='{range .items[{asterisk}]} ...')*
```
**Issues:**
- Adds complexity with attribute definitions
- Makes source less intuitive to read and edit
- Requires consistent attribute management across documents

## Possible Solution

To avoid parsing conflicts and achieve the desired bold formatting, a possible solution might be to specify a role in the source, `role=command`, and combine that with CCS. 

```asciidoc
[source,terminal,role=command]
----
$ for I in $(oc get ns -o jsonpath='{range .items[*]} {.metadata.name}{"\n"} {end}'); \
    do oc delete pods --all -n $I; \
    sleep 1; \
    done
----
```

Then apply styling via CSS:
```css
.command {
    font-weight: bold;
    background: #f5f5f5;
[...]

PR #543, opened 2025-09-18
[GH#535]: Remove asterisk formatting guidance for commands

rolfedh:
This PR removes the requirement to enclose commands in asterisks for bold formatting. This formatting breaks commands that contain wildcard characters (`*`), making them non-functional when copied and pasted.

**Issue:** #535

**Impact:** This bug must be fixed before the organization implements the new guidance for separating commands from output, as it would otherwise affect a significant number of command examples.

**Next steps:** The organization should determine whether bold formatting for commands remains necessary now that commands are visually separated from output. If bold formatting is still desired, we should evaluate alternative implementation methods (such as CSS) that won't interfere with command syntax.

@IngridT1


IngridT1:
@bergerhoffer @rolfedh I'll take care of those next steps. I'll start a conversation on Slack.

IngridT1 (approved):
I'm approving this change. I'll also put in a PR to remove the guidance for boldfacing commands.

PR #549, opened 2025-10-13
Improving the adoc example for Commands in code blocks

jherrman:
In the current rendering of https://redhat-documentation.github.io/supplementary-style-guide/#commands-in-code-blocks , the "Example asciidoc" displays as 3 separate codeblocks, which is confusing and does not reflect the contents of the codeblocks being used sequentially in the same adoc module.

This rewrite aims to fix that by ensuring that there is just one code block instead of 3.


IngridT1 (approved):
LGTM
But I would like to get one more approval before I merge.

PR #555, opened 2025-11-18
Adding a new term: key pair

jseseCCS:
Adding a new term: key pair. I've had to use it more than once and I've seen it spelled as two words and one and I can never remember which is correct. Not addressed in IBMSG.

No need to add if folks think it's unnecessary.


IngridT1 (changes requested):
This addition to the style guide looks good to me. Please make this small change.

PR #559, opened 2025-11-25
Test

aireilly:
rh-pre-commit.version: 2.3.2
rh-pre-commit.check-secrets: ENABLED

Issue:

Additional information:


bburt-rh:
@aireilly the preview link works now. The comment gets added before the build completes, so maybe we need to fix that timing

PR #581, opened 2026-03-12
We do not include OpenShift CRDs as Style entries

aireilly:
We do not include OpenShift CRDs as Style Guide entries.

### DIFF
--- a/supplementary_style_guide/glossary_terms_conventions/general_conventions/i.adoc
+++ b/supplementary_style_guide/glossary_terms_conventions/general_conventions/i.adoc
@@ -381,18 +381,6 @@ There is no functional difference between the first server that was installed an
 
 *See also*:
 
-[[inferenceservice]]
-==== image:images/yes.png[yes] InferenceService (noun)
-*Description*: In Red Hat OpenShift AI, this is the custom resource definition (CRD) used to create the `InferenceService` object. When referring to the CRD name, use `InferenceService` in monospace. 
-          
-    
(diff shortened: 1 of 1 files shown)


IngridT1:
@aireilly I think we should retain this entry for disambiguation reasons.

aireilly:
> @aireilly I think we should retain this entry for disambiguation reasons.

Disambiguation from what though? There are no other OpenShift CRDs in the Style Guide, and this entry even includes the exact same term in the "do not use" section.

IngridT1:
_InferenceService (noun)_ should remain in the SSG because the entry explicitly says not to use _inference serving_ for the _InferenceService_ CRD. That's the disambiguation that I meant. I think that we can include a CRD if we're having problems with people using a known term incorrectly to refer to it. I think that's why @kelbrown20 included it in the initial PR.

aireilly:
> _InferenceService (noun)_ should remain in the SSG because the entry explicitly says not to use _inference serving_ for the _InferenceService_ CRD. That's the disambiguation that I meant. I think that we can include a CRD if we're having problems with people using a known term incorrectly to refer to it. I think that's why @kelbrown20 included it in the initial PR.

The InferenceService entry treats a Kubernetes CRD (`InferenceService`) as if it were just a misused generic term. It is not. Inference serving is a general concept, InferenceService is a very specific API resource name.

IngridT1:
I still don't think it's a problem to include it.

Issue #564, opened 2025-12-05
Contradiction in the Release Notes section about the use of "now"


mramendi:
### Severity

Major (significant impact, but workaround exists)

### Priority

P2

### Affected page(s)

https://redhat-documentation.github.io/supplementary-style-guide/#tenses-in-release-notes

### Describe the bug, along with repro steps

https://redhat-documentation.github.io/supplementary-style-guide/#tenses-in-release-notes says: "Do not use "now" to refer to the state after the update." In the very next section offers the following example heading: "nvme-cli and cryptsetup are now available for Opal automation on NVMe SEDs".

### Describe the proposed fix

Either drop the restriction on "now" or fix the example to not use "now"

IngridT1:
@mramendi Merged. Thank you!

PR #582, opened 2026-03-30
fix(release-notes): remove contradictory "now" from examples

IngridT1:
Align release note examples with the rule to avoid "now" for the post-update state (issue #564). Rephrase the katello-agent deprecation example and CCFR template result lines for consistency.

Made-with: Cursor

Issue:

Additional information:


sbmetz (approved):
LGTM!

PR #585, opened 2026-04-14
Add "change" to glossary

apinnick:
Issue: None

Additional information:

### DIFF
--- a/supplementary_style_guide/glossary_terms_conventions/general_conventions/c.adoc
+++ b/supplementary_style_guide/glossary_terms_conventions/general_conventions/c.adoc
@@ -266,6 +266,17 @@
 
 *See also*:
 
+[[change]]
+==== image:images/caution.png[caution] change (verb)
+*Description*: Use with caution. Instead, use a descriptive verb such as "edit", "update", or "modify" unless the context is changing directories or passwords.
+
+*Use it*: with caution
+
+[.vale-ignore]
+*Incorrect forms*: 
+
+*See also*:
+
 [[cidr]]
 ==== image:images/yes.png[yes] CIDR (noun)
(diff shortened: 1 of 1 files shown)


IngridT1:
Could you add more context for why you feel this addition is necessary, @apinnick ?

apinnick:
> Could you add more context for why you feel this addition is necessary, @apinnick ?

@IngridT1 The verb "change" is not an industry standard except in contexts such as changing passwords or directories (even then, we tend to use "navigate to", although `cd` command means "change directory"). ISG does not use "change" or "changing" except in the context of changing passwords or default settings.

I noticed "Changing ..." in [Importing content](https://docs.redhat.com/en/documentation/red_hat_satellite/6.18/html/managing_content/importing_content_content-management), Satellite _Managing content_ guide. It is only used in modules that were written more than 4 years ago. It looks strange and vague when the rest of the document uses more descriptive verbs such as "Creating", "Adding", "Removing", etc. I don't think "Changing" should be forbidden, just used sparingly. 

The Satellite writers used descriptive verbs when they wrote abstracts for DITA migration (see screenshot). However, it was too much trouble to change the titles because the upstream repo require anchor IDs to match titles, which would have resulted in a lot of broken links and xrefs.

image

IngridT1:
I don't think this is a necessary addition, but I would love for @jherrman @bergerhoffer @lburcher7 @bburt-rh to weigh in. They might not agree with me.

jherrman:
Personally, I don't see anything wrong with using the term "change" in situations where you replace an existing setting/item with a partially or completely different one. In that way, the change in "changing the default download policy" is semantically very close to "changing the password".

Still, Avital has a point that if you can use a more specific verb like "create", "add", or "remove", then "change" is unnecessarily vague. Nevertheless, I'm not really sure how much that could be an issue across our doc sets. Still, it shouldn't hurt to have a shortish mention about avoiding the term where possible, as a best practice.

PR #590, opened 2026-05-06
Improve markdown output quality

sslocket:
Replace the sed-based cleanup in the deploy workflow with two Node.js scripts that fix heading hierarchy, broken anchor links, image artifacts, and empty glossary fields. No new dependencies are introduced — Node.js is already used for downdoc.


sslocket:
@IngridT1 @aireilly

aireilly (commented):
Over 400 lines of code! Could we slim this down? We don't need all the comments for one thing.

sslocket:
@aireilly I committed an abridged version.

aireilly (approved):
/lgtm

